PR #62, opened 2026-01-06
Fix critical jspdf vuln and clean up pre-commit hooks

RinZ27:
Found some critical issues during a recent audit that we really shouldn't leave hanging. 

The most urgent fix is for `jspdf`—previous versions had a nasty Path Traversal bug (GHSA-f8cm-6447-x5h2) that's now patched by bumping to v4. I also took the opportunity to clean up our pre-commit pipeline. `ruff` was configured to essentially ignore failures, which defeats the purpose of having it, and ESLint was sitting idle. I've re-enabled both to ensure we're catching broken code before it hits the repo.

Finally, I updated `vitepress` in the docs to clear out some moderate CVEs affecting the development server.

Make sure to run a fresh `npm install` in `/web` and `/docs` after pulling this.

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -372,9 +372,9 @@
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.21.5",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.21.5.tgz",
-      "integrity": "sha512-1SDgH6ZSPTlggy1yI6+Dbkiz8xzpHJEVAlF/AM1tHPLsf5STom9rwtjE4hKAF20FfXXNTFqEYXyJNWh1GiZedQ==",
+      "version": "0.25.12",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.25.12.tgz",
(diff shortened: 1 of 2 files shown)


scrrlt:
i’d recommend splitting this into 2 pr’s — so the urgent security patch can be merged asap and in isolation, without complicating any rollback if it becomes necessary.

minor nitpick — jspdf v4 introduced some breaking changes around plugin loading and constructor options. worth double‑checking that nothing in /web relies on deprecated behaviour.

re: --exit-zero — correct for ci, but if there are existing lint violations in the repo it will block contributor workflow in pre‑commit. not sure if you’ve run a repo‑wide lint pass. interim recommendation: add --fix-only or a restricted --select set so the hook doesn’t block on warnings until the repo is clean.

eslint: the pr description says it was re‑enabled, but the hook is still commented out. if the intention is to actually re‑enable it, make sure:

- web/.eslintrc exists and matches the version you pinned
- next.js eslint config is compatible with the version bump
- the hook won’t block contributors until the codebase is lint‑clean
- 
if the intention is to disable eslint for now, the pr description should reflect that.
there are still some unused python imports in display_manager.py (console, layout, live, richhandler, panel, table, text). worth pruning if they’re genuinely unused.

token tracker — you removed completion_cost; double‑check that no downstream code still references it.

nitpick: instead of
sys.path.insert(0, str(project_root))
prefer setting PYTHONPATH=. or using a proper conftest.py fixture. avoids masking import issues.

package‑lock changes look mostly like regeneration, but just confirm npm version consistency so we don’t introduce accidental churn.

overall: i’d keep security patches isolated and fast‑tracked, and move tooling changes into a separate pr for easier reversibility.

pancacake:
thanks for your contribution!

PR #71, opened 2026-01-07
feat/gracefully disable RagAnything when absent and tighten Python dependency handling

scrrlt:
## Summary
- declare every backend dependency in `pyproject.toml`, cap Python to <3.14,
  disable Poetry’s package mode so installs succeed, and add a `dev` extra
  (`pytest>=8.3,<9.0`) — then regenerate `poetry.lock` and align
  `requirements.txt` to match those ranges
- remove RagAnything from the default dependency list, introduce a shared
  `src/knowledge/raganything_loader.py`, and refactor the ingestion scripts
  plus `rag_tool.py` to log a clear warning and skip MinerU-backed flows when
  the optional dependency is missing instead of crashing the entire app
- document the new workflow in `README.md`, explicitly outlining the dependency
  layers (core backend via `poetry install --no-root`, dev extras via
  `--extras dev`, optional RagAnything/MinerU install, and frontend npm setup),
  and update `tests/core/test_prompt_manager.py` so caching behavior remains
  testable without RagAnything on the path

## Root Cause
Fresh environments that skipped RagAnything (and therefore MinerU) hit a
`ModuleNotFoundError` during import because our ingestion modules pulled
RagAnything at module load. Once that happened FastAPI never even started,
rendering the whole application unusable for those that only needed the core
features.

## Fix
Treat RagAnything as an optional capability at the application boundary:
imports are now wrapped, missing dependencies emit human-friendly logs, and
the rest of the system stays operational. Advanced PDF/OCR ingestion still
requires installing RagAnything + MinerU manually, and the README now calls
that out. Fully decoupling MinerU inside RagAnything would still require
upstream changes (feature flags, guarded imports, fallback parsers), but this
PR ensures the host app degrades gracefully in the meantime.

## Dependency Model (for reviewers)
- **Core backend** – `poetry install --no-root`, using the ranges declared in
  `pyproject.toml`.
- **Developer extras** – `poetry install --no-root --extras dev`, which brings
  in pytest and the rest of the test/lint tooling.
- **RagAnything/MinerU (optional)** – Not installed by default. Teams wanting
  PDF/OCR ingestion should install RagAnything alongside this repo and follow
  its instructions for MinerU.
- **Frontend** – `npm install --prefix web` remains unchanged.

## Testing
- py -3.13 -m poetry run ruff check src/knowledge/add_documents.py src/knowledge/initializer.py src/tools/rag_tool.py
- py -3.13 -m poetry run pytest tests/core/test_prompt_manager.py

## Notes
`src/knowledge/raganything_loader.py` now bundles the lazy import logic for the
ingestion scripts and RAG tool. If we later fork RagAnything to make MinerU
optional upstream, we can swap out the helper’s internals in one place without
touching the rest of the application.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,8 +44,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
-          pip install pytest pytest-asyncio pytest-cov
+          pip install -e .[dev]
+          pip install pytest-asyncio pytest-cov
+
+      - name: Verify dependencies
+        run: |
+          pip check
 
       - name: Run tests with coverage
         run: |
@@ -80,7 +84,11 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install -e .[dev]
+
+      - name: Verify dependencies
+        run: |
+          pip check
 
       - name: Check core module imports
         run: |
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -35,6 +35,11 @@ python scripts/install_all.py
 # Or Install Dependencies Manually
 pip install -r requirements.txt
 npm install
+
+# Verify Python environment
+pip install -e .[dev]
+pip check
+pytest -q
 ```
 
 ## Step 3: Set Up Environment Variables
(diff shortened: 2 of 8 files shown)


scrrlt:
> Thanks for your contribution! Will take a detailed look. By the way, do you want to join our collaborator team? We'll timely release some TODOs for core collaborators to choose and complete. I remember that you've contributed before~

sure thang

PR #147, opened 2026-01-18
feat: Increase LLM retry attempts and delays for enhanced robustness with complex agents.

tan-huynh:


pancacake:
Thanks for your contribution !

PR #196, opened 2026-02-12
Restrict code execution scope and enforce import whitelist

RinZ27:
Noticed the code execution tool was letting LLM-generated scripts roam around the project root and access all environment variables. To fix this, I restricted the workspace to the user data directory and stripped sensitive keys from the environment before execution.

Enforcing a default whitelist for imports via `ImportGuard` ensures that only safe libraries like numpy or matplotlib are used, even if the caller forgets to specify constraints. I've also updated the solver and research pipelines to leverage these restricted settings by default.

Confirmed that core functionalities like plotting and calculations still work fine under these tighter constraints. This should stop any potential RCE from escaping the sandbox or exfiltrating API keys.


pancacake:
yea that's a good one. Thanks!

PR #272, opened 2026-04-09
fix: use lowercased filename in mimetypes.guess_type() for consistent MIME validation

### LINKED ISSUE #271: bug: mimetypes.guess_type() uses original case filename causing inconsistent MIME validation in DocumentValidator
### Description
In `deeptutor/utils/document_validator.py`, the `validate_upload_safety()` method has an inconsistency in how it handles filenames for extension vs MIME type validation.

The file extension `ext` is extracted from the **lowercased** filename:
```python
_, ext = os.path.splitext(filename.lower())
```

But MIME type guessing uses the **original** filename (preserving case):
```python
guessed_mime, _ = mimetypes.guess_type(filename)
```

### Impact
On Linux, `mimetypes.guess_type()` is case-sensitive. This means:
- A file named `report.PDF` passes the extension check (`ext = ".pdf"` ✓)
- But `mimetypes.guess_type("report.PDF")` may return `None` on Linux
- This causes a `ValueError` about MIME type validation failure for an otherwise valid upload

Meanwhile on Windows, `mimetypes.guess_type()` is case-insensitive and works fine — making this a platform-dependent bug that is harder to reproduce.

### Steps to Reproduce
```python
from deeptutor.utils.document_validator import DocumentValidator
# On Linux with a file that has uppercase extension:
DocumentValidator.validate_upload_safety("report.PDF", 1024)
# May raise: ValueError: MIME type validation failed: None. File may be malicious or corrupted.
```

### Expected Behavior
Files with valid extensions but mixed-case names (e.g., `report.PDF`, `notes.TXT`) should be accepted consistently regardless of OS.

### Fix
Pass the lowercased filename to `mimetypes.guess_type()` to ensure consistency with the extension check:
```python
guessed_mime, _ = mimetypes.guess_type(filename.lower())
```

### Environment
- Python 3.10+
- Linux (case-sensitive filesystem)

kuishou68:
Closes #271

## Problem
`validate_upload_safety()` in `DocumentValidator` extracts the file extension from the **lowercased** filename:
```python
_, ext = os.path.splitext(filename.lower())
```

But MIME type guessing uses the **original** filename (preserving the original case):
```python
guessed_mime, _ = mimetypes.guess_type(filename)
```

On Linux (case-sensitive), `mimetypes.guess_type("report.PDF")` returns `None`, causing a `ValueError` for valid files with uppercase extensions. This is a platform-dependent bug — it works on Windows (case-insensitive MIME lookup) but fails on Linux.

## Fix
Pass `filename.lower()` to `mimetypes.guess_type()` to match the extension check:
```python
guessed_mime, _ = mimetypes.guess_type(filename.lower())
```

This ensures files like `report.PDF`, `notes.TXT`, or `data.JSON` are accepted consistently on all platforms.

### DIFF
--- a/deeptutor/utils/document_validator.py
+++ b/deeptutor/utils/document_validator.py
@@ -106,7 +106,7 @@ class DocumentValidator:
             )
 
         # Additional MIME type validation for security
-        guessed_mime, _ = mimetypes.guess_type(filename)
+        guessed_mime, _ = mimetypes.guess_type(filename.lower())
         if guessed_mime and guessed_mime not in DocumentValidator.ALLOWED_MIME_TYPES:
             raise ValueError(
                 f"MIME type validation failed: {guessed_mime}. File may be malicious or corrupted."


pancacake:
Thanks for pointing that out!

PR #306, opened 2026-04-13
fix: pass extra_headers to llm_complete in config test runner

OldSuns:
## Problem

When users configure custom `extra_headers` (e.g., custom `User-Agent`) in the model settings page, the **connection test** still sends the default SDK headers (e.g., `user-agent: AsyncOpenAI/Python 2.31.0`) instead of the user-configured values.

This happens because `_test_llm` in `ConfigTestRunner` creates an `LLMConfig` with `extra_headers` correctly, but never passes it to the `llm_complete()` function call.

## Root Cause

`deeptutor/services/config/test_runner.py` — `llm_complete()` invocation missing `extra_headers` kwarg.

## Fix

One-line change: add `extra_headers=llm_config.extra_headers` to the `llm_complete()` call in `_test_llm`.

```diff
+            extra_headers=llm_config.extra_headers,
```

## Impact

- **Users with custom headers**: Model connection test now correctly sends user-configured headers
- **Regular users (no extra_headers)**: No change — `{}` is falsy and skipped by `default_headers.update()`
- Fully backward compatible

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,17 +40,6 @@
 
 > **[2026.4.4]** [v1.0.0-beta.1](https://github.com/HKUDS/DeepTutor/releases/tag/v1.0.0-beta.1) — Agent-native architecture rewrite (～200k lines) with two-layer plugin model (Tools + Capabilities), CLI & SDK entry points, TutorBot multi-channel bot agent, Co-Writer, Guided Learning, and persistent memory.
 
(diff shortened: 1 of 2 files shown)


OldSuns:
Closing to rebase on dev branch properly. Will reopen with correct base.

PR #345, opened 2026-04-20
fix: disable json_object response_format for gemma models

### LINKED ISSUE #344: [Bug]: response_format.type

### Describe the bug

I successfully deployed the model ` gemma-4-e2b ` in LM Studio, and the chat is working properly.
I can't make visualization, error: 
` openai.BadRequestError: Error code: 400 - {'error': "'response_format.type' must be 'json_schema' or 'text'"} `
I'm looking at the logs, I saw that the request for the model is:
` "response_format": {
"type": "json_object"
} `
I haven't checked the other functions, maybe they have this problem too.

### Steps to reproduce

_No response_

### Expected Behavior

The visualization function is working properly

### Related Module

API/Backend

### Configuration Used

_No response_

### Logs and screenshots

_No response_

### Additional Information

- DeepTutor Version:
- Operating System:
- Python Version:
- Node.js Version:
- Browser (if applicable):
- Related Issues:

octo-patch:
Fixes #344

## Problem

When using LM Studio with `gemma` models (e.g. `gemma-4-e2b`), the
visualization and other JSON-structured features fail with:

```
openai.BadRequestError: Error code: 400 - {'error': "'response_format.type' must be 'json_schema' or 'text'"}
```

The root cause: the `lm_studio` binding has `supports_response_format: True`
in `PROVIDER_CAPABILITIES`, so the code sends `response_format={"type": "json_object"}`.
However, newer Gemma models only accept `json_schema` or `text` — they do not
support the legacy `json_object` type.

## Solution

Add `"supports_response_format": False` to the existing `"gemma"` entry in
`MODEL_OVERRIDES` inside `deeptutor/services/llm/capabilities.py`.

When `supports_response_format` returns `False`, the code omits the
`response_format` parameter entirely. The existing `extract_json_object`
utilities in the `visualize` and `math-animator` agents already parse
structured JSON from plain-text responses, so all callers work correctly
without `response_format` being set.

This is the same pattern already used for DeepSeek and Anthropic models.

## Testing

- Direct Python assertions covering `gemma-4-e2b`, `gemma-3-4b`, `gemma-2-9b` all return `False`
- Non-gemma LM Studio models (`mistral-7b`, `llama-3`) continue to return `True`
- Added `test_gemma_response_format_disabled` to `tests/services/llm/test_capabilities.py`

### DIFF
--- a/deeptutor/services/llm/capabilities.py
+++ b/deeptutor/services/llm/capabilities.py
@@ -192,7 +192,7 @@ MODEL_OVERRIDES: dict[str, dict[str, object]] = {
     "claude-3": {"supports_vision": True},
     "claude-4": {"supports_vision": True},
     "gemini": {"supports_vision": True},
-    "gemma": {"supports_vision": False},
+    "gemma": {"supports_vision": False, "supports_response_format": False},
     "llava": {"supports_vision": True},
     "bakllava": {"supports_vision": True},
     "moondream": {"supports_vision": True},
--- a/tests/services/llm/test_capabilities.py
+++ b/tests/services/llm/test_capabilities.py
@@ -14,6 +14,20 @@ def test_model_override_capability() -> None:
     assert has_thinking_tags("openai", "deepseek-reasoner") is True
 
 
+def test_gemma_response_format_disabled() -> None:
+    """Gemma models do not support json_object response_format (only json_schema/text).
+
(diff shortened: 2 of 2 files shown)


mr-Lime197:
this problem persists with qwen-like architectures.

pancacake:
Thanks for your contribution!

PR #349, opened 2026-04-20
feat(chat): regenerate last response (CLI, WebSocket, Web UI)

DarkGenius:
## Summary

Adds a "regenerate last response" capability that re-runs the prior user message in a session as a fresh turn **without** duplicating the user row. Exposed across all three entry points:

- **WebSocket** — new `type: "regenerate"` message accepting an optional `overrides` field; mirrors the `start_turn` flow including `subscribe_turn`. Errors `regenerate_busy` / `nothing_to_regenerate` surface as non-fatal `error` events.
- **CLI** — `/regenerate` (alias `/retry`) inside `deeptutor chat`, backed by a shared `regenerate_and_render` helper.
- **Web UI** — per-message Regenerate button (`RefreshCcw` icon) sits next to Copy on the **last** assistant message. Uses a new `regenerateLastMessage()` action that pops the trailing assistant optimistically and sends the new WS message. Replaces the previous Retry button (which depended on an in-memory `requestSnapshot` and silently disappeared after page reload), so regeneration now also works for **historical sessions** hydrated from the store.

## Design

- `SQLiteSessionStore.delete_message` and `get_last_message` helpers added for tail rollback (append-only invariant preserved — only the trailing row is ever removed, so `summary_up_to_msg_id` is unaffected).
- `TurnRuntimeManager.regenerate_last_turn(session_id, overrides=None)`:
  - rejects when an active turn is running (`regenerate_busy`),
  - rejects when no prior user message exists (`nothing_to_regenerate`),
  - deletes the trailing assistant (if any),
  - rebuilds the payload from session preferences + the stored user message (including attachments),
  - sets `_persist_user_message=False` and `_regenerate=True`, then delegates to the regular `start_turn` to keep a single execution path.
- New `runtime_only_keys` plus SESSION-event metadata (`regenerated_from_message_id`, `superseded_turn_id`, `regenerate=true`) so clients can reconcile the UI without a schema migration. Old turn rows are kept untouched (no destructive history mutation).
- `_run_turn` skips `memory_service.refresh_from_turn` on regeneration to avoid double-updating long-term memory for the same user prompt.

## Operational improvement bundled in

`AgenticChatPipeline._stage_responding` now warns and emits a non-terminal `error` event when the LLM returns an empty stream (zero non-empty chunks). The warning includes model, chunk count, prompt char size, and observation size so empty assistant messages can be diagnosed from logs and the Trace panel without further code instrumentation. This is independent of regenerate but materially improves UX when a model returns nothing — the user now sees a clear hint to hit Regenerate instead of staring at a silent empty bubble.

## Test plan

`tests/services/session/test_regenerate.py` (18 tests):

- `_extract_regenerate_flag` parsing.
- Store: `delete_message` removes only the targeted row; `get_last_message` with/without role filter and on empty sessions.
- Runtime unit tests with patched `start_turn`:
  - last is `assistant` → assistant deleted, payload replays user content / capability / tools / kb / language / attachments;
  - last is `user` → no deletion, new turn proceeds;
  - empty / missing session → `nothing_to_regenerate`;
  - active running turn → `regenerate_busy`;
  - overrides take precedence yet runtime flags are still injected.
- End-to-end integration test wiring a fake orchestrator + tracking memory service: original user is preserved, prior assistant is replaced, the new SESSION event carries `regenerate=true` and `regenerated_from_message_id`, and `refresh_from_turn` is **not** invoked for the regenerated turn.

All 18 + neighbouring tests in `tests/services/session/`, `tests/api/test_unified_ws_turn_runtime.py`, `tests/cli/test_chat_cli.py`, `tests/services/test_app_facade.py` pass locally.

### Manual smoke

- Web UI: open any chat session (live or historical), click Regenerate under the last assistant message → optimistic deletion + new generation; user message preserved in the DB.
- CLI: `deeptutor chat`, send a message, then `/regenerate` → re-runs the previous prompt with the same preferences; `/retry` is an alias.
- WS: client sends `{ "type": "regenerate", "session_id": "..." }` → `subscribe_turn` for the new turn id; `regenerate_busy` / `nothing_to_regenerate` returned as `error` event with `metadata.reason`.

## Notes

- No schema migration required.
- No new Python or npm dependencies.
- AGENTS.md and the `unified_ws.py` module docstring updated to document the new CLI command and WS message type.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,7 @@ deeptutor run deep_question "Linear algebra" --config num_questions=5
 
 # Interactive REPL
 deeptutor chat
+# (inside the REPL: /regenerate or /retry re-runs the last user message)
 
 # Knowledge bases
 deeptutor kb list
--- a/deeptutor/agents/chat/agentic_pipeline.py
+++ b/deeptutor/agents/chat/agentic_pipeline.py
@@ -391,7 +391,49 @@ class AgenticChatPipeline:
                     {"trace_kind": "call_status", "call_state": "complete"},
                 ),
             )
-            return clean_thinking_tags("".join(chunks), self.binding, self.model), trace_meta
+            final_response = clean_thinking_tags("".join(chunks), self.binding, self.model)
+            if not final_response.strip():
+                # The provider returned an empty stream (zero non-empty chunks)
+                # or only whitespace. This typically means: model hit a token
+                # limit, was filtered, or treated the observation as the final
+                # answer. Surface a non-terminal warning so the operator sees
+                # the cause in logs and the UI can hint a Regenerate.
+                prompt_chars = sum(len(str(m.get("content") or "")) for m in messages)
+                logger.warning(
+                    "[%s] responding stage returned empty response "
+                    "(model=%s, chunks=%d, prompt_chars=%d, max_tokens=%d, observation_chars=%d)",
+                    trace_meta.get("call_id"),
+                    self.model,
+                    len(chunks),
+                    prompt_chars,
+                    self._chat_limits.responding,
+                    len(observation or ""),
+                )
+                await stream.error(
+                    self._t(
+                        "notices.empty_response",
+                        default=(
+                            "The model returned an empty response. "
+                            "Try Regenerate or rephrase the question."
+                        ),
+                    ),
+                    source="chat",
+                    stage="responding",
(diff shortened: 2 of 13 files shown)


pancacake:
Thanks for your contribution! Just done what i'm going to do lol

PR #390, opened 2026-04-25
[codex] Normalize Azure OpenAI max token aliases

54sun:
## Summary
- normalize max_completion_tokens to max_tokens before Azure OpenAI provider execution
- add regression coverage for Azure OpenAI factory calls

## Root cause
Newer model helpers emit max_completion_tokens for gpt-5.* models, but DeepTutor's Azure Responses provider accepts max_tokens and maps that internally to max_output_tokens. The alias leaked through and caused Azure OpenAI calls to fail before reaching the API.

## Validation
- python -m py_compile deeptutor/services/llm/factory.py tests/services/llm/test_factory_provider_exec.py
- live Docker backend LLM and embedding connection tests passed

### DIFF
--- a/deeptutor/services/llm/factory.py
+++ b/deeptutor/services/llm/factory.py
@@ -272,6 +272,8 @@ def _sanitize_call_kwargs(
 
     if not supports_response_format(binding, model):
         extra_kwargs.pop("response_format", None)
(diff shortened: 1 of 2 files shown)


pancacake:
fixed through pr#438.

PR #428, opened 2026-04-30
fix: fallback content to reasoning_content when DeepSeek returns empty content field

Starfie1d1272:
## Problem

DeepSeek v4-flash and v4-pro default to thinking mode enabled (per [API docs](https://api-docs.deepseek.com/guides/thinking_mode)), returning responses in `reasoning_content` while leaving `content` empty. This caused cascade failures across multiple layers:

| Symptom | Root cause |
|---|---|
| `LLM returned empty response` | `_parse()` / `_parse_chunks()` only fell back to `m.reasoning` (DashScope), not `m.reasoning_content` (DeepSeek) |
| `No JSON object found` in book blocks | `CodeGenerator` / `FlashCardsGenerator` / `TimelineGenerator` / `DeepDiveGenerator` used fragile `json.loads()` directly |
| Streaming empty response | `factory.py` wrapped reasoning in `<think>` tags stripped by `clean_thinking_tags()` |
| `'list' object has no attribute 'get'` | `IdeaAgent` didn't guard against array-type JSON responses |
| `LLM_REASONING_EFFORT` env var silently ignored | resolver path takes priority over `.env`; env override not applied |

### 1. Content fallback + precedence alignment
- `_parse()` / `_parse_chunks()` in both `provider_core` and `tutorbot`: fall back `reasoning_content` → `content`, check `reasoning_content` before `reasoning`

### 2. Streaming path
- `factory.py` `_runner`: when only reasoning chunks are emitted (no `_on_content_delta`), emit `response.content` as fallback

### 3. Book block JSON parsing
- `code.py`, `flash_cards.py`, `timeline.py`, `deep_dive.py`: replace raw `json.loads()` → `parse_json_response()` which handles markdown fences, preamble, and `json_repair` fallback

### 4. `LLM_REASONING_EFFORT` env var
- `config.py`: read from env and apply as override regardless of resolver path
- `factory.py`: pass `config.reasoning_effort` to `chat_with_retry` / `chat_stream_with_retry`
- For DeepSeek: `high`/`max` enables thinking, `minimal` sends `thinking: {type: "disabled"}`, empty → auto-detect
- `.env.example` / `.env.example_CN`: documented

### 5. idea_agent robustness
- Guard against JSON array and non-dict responses

## Scope note

This PR is a **bug fix** — it prevents crashes and empty responses when thinking mode is active. Full thinking mode support (leveraging `reasoning_content` to improve generation quality, token budget allocation, animation pipeline adaptation) is tracked in #430.

Also tracked in #430: book generation UX improvements (force regenerate, failure diagnostics, prompt leakage prevention, i18n).

## Test plan
- [ ] `deepseek-v4-flash` + `LLM_REASONING_EFFORT=minimal`: all book blocks generate without errors
- [ ] `deepseek-v4-flash` + `LLM_REASONING_EFFORT=high`: reasoning_content falls back correctly, no empty responses
- [ ] `deepseek-chat` / `gpt-4o` regression: normal operation unchanged
- [ ] Idea generation handles malformed LLM JSON


pancacake:
Thanks for your contribution!

PR #438, opened 2026-05-03
fix(llm): map max_completion_tokens to max_output_tokens for Responses API

### LINKED ISSUE #437: [Bug]: AsyncResponses.create() got an unexpected keyword argument 'max_completion_tokens' when using OpenAI gpt-5.x models

### Describe the bug

gpt-5.x models fail under latest release.  Using gpt-4o-mini or other models works fine.

### Steps to reproduce

Steps to reproduce:
Set LLM_BINDING = openai
Set LLM_MODEL = gpt-5.5
Launch official docker container for v1.3.5
Try chatting, or generating a quiz.  Nothing will come back.

### Expected Behavior

Should get a response.

### Related Module

Dashboard

### Configuration Used

Use docker-compose.yaml example from the repo, unchanged.

### Logs and screenshots

_No response_

### Additional Information

- DeepTutor Version: 1.3.5
- Official Docker Pre-built GHCR Image

The observed behavior looks similar to https://github.com/HKUDS/DeepTutor/issues/54 but I don't know if the underlying cause is the same.

truffle-dev:
## What

When the Responses API path is selected for newer OpenAI models (gpt-5.x, o1, o3, o4) and an extra kwarg of \`max_completion_tokens\` flows in, the OpenAI SDK raises \`TypeError\` from \`responses.create()\` before any HTTP request leaves the client. Closes #437.

## Why

\`get_token_limit_kwargs(model, n)\` returns \`{"max_completion_tokens": n}\` for newer models (the Chat Completions name). Both \`OpenAICompatProvider\` and \`AzureOpenAIProvider\` route those kwargs through \`client.responses.create()\` when \`_should_use_responses_api()\` matches, but the Responses API only accepts \`max_output_tokens\`. The SDK rejects the unknown kwarg with \`TypeError\` inside the client, and \`_should_fallback_from_responses_error()\` only catches HTTP errors with \`status_code\` — so the call is never retried via Chat Completions.

The reporter in #437 hit this with \`LLM_BINDING = openai\` and \`LLM_MODEL = gpt-5.5\` on v1.3.5, which routes through \`OpenAICompatProvider\`.

## How

Add \`adapt_chat_kwargs_to_responses()\` to \`provider_core/openai_responses/converters.py\` and use it at all four merge sites: \`chat\` and \`chat_stream\` on both \`OpenAICompatProvider\` (lines 698, 781) and \`AzureOpenAIProvider\` (lines 126, 154).

The helper:
- preserves the existing \`None\`-drop semantics of the previous dict-comprehension merge,
- maps \`max_completion_tokens\` → \`max_output_tokens\`,
- only applies the alias when \`max_output_tokens\` is not explicitly set by the caller,
- does not mutate the input.

## Tests

\`tests/services/llm/test_openai_responses_converters.py\` covers seven cases: passthrough, None-drop, the rename, None for the alias, explicit-name precedence, empty input, and input non-mutation. \`pytest tests/services/llm/test_openai_responses_converters.py -v\` is green; the rest of \`tests/services/llm/\` and \`tests/services/config/test_llm_probe_config.py\` pass with no new failures from this branch (75/76; the one unrelated failure is a missing \`data/user/settings/agents.yaml\` fixture).

## Note on #390

There is an existing open PR (#390, "[codex] Normalize Azure OpenAI max token aliases") that addresses a related problem at the factory layer for Azure only by translating to \`max_tokens\`. The reproduction in #437 uses \`LLM_BINDING = openai\`, which goes through \`OpenAICompatProvider\` and is outside that PR's scope. Happy to defer if a different shape is preferred.

### DIFF
--- a/deeptutor/services/llm/provider_core/azure_openai_provider.py
+++ b/deeptutor/services/llm/provider_core/azure_openai_provider.py
@@ -11,6 +11,7 @@ from openai import AsyncOpenAI
 
 from deeptutor.services.llm.provider_core.base import LLMProvider, LLMResponse
 from deeptutor.services.llm.provider_core.openai_responses import (
+    adapt_chat_kwargs_to_responses,
     consume_sdk_stream,
     convert_messages,
     convert_tools,
@@ -122,7 +123,7 @@ class AzureOpenAIProvider(LLMProvider):
             reasoning_effort,
             tool_choice,
         )
-        body.update({k: v for k, v in extra_kwargs.items() if v is not None})
+        body.update(adapt_chat_kwargs_to_responses(extra_kwargs))
         try:
             return parse_response_output(await self._client.responses.create(**body))
         except Exception as exc:
@@ -150,7 +151,7 @@ class AzureOpenAIProvider(LLMProvider):
             reasoning_effort,
             tool_choice,
         )
-        body.update({k: v for k, v in extra_kwargs.items() if v is not None})
+        body.update(adapt_chat_kwargs_to_responses(extra_kwargs))
         body["stream"] = True
         idle_timeout_s = 90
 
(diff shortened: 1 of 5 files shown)


pancacake:
Thanks for your contribution!

PR #485, opened 2026-05-14
fix(auth): make require_auth async so the user ContextVar reaches the endpoint

### LINKED ISSUE #481: [Bug] Non-admin users get 404 on all session requests in Docker multi-user deployment

### Describe the bug

When AUTH_ENABLED=true, non-admin users can log in successfully but cannot chat. Every session request returns 404 and the per-user workspace directory is never created on disk.

### Steps to reproduce

1. Deploy via docker-compose.ghcr.yml with AUTH_ENABLED=true
2. Register first user via /register → becomes admin
3. Create a second user via /admin/users
4. Assign LLM model grant to the second user
5. Log in as the second user in an incognito window
6. Attempt to send a chat message

### Expected Behavior

Non-admin user gets their own isolated workspace under multi-user/<uid>/ and can chat normally.

### Related Module

Dashboard

### Configuration Used

AUTH_ENABLED=true
NEXT_PUBLIC_AUTH_ENABLED=true
AUTH_SECRET=<set>
LLM_BINDING=openai
LLM_MODEL=deepseek-v4-flash
LLM_HOST=https://api.deepseek.com/v1
EMBEDDING_BINDING=ollama
EMBEDDING_MODEL=nomic-embed-text

### Logs and screenshots

Backend logs:
GET /api/v1/sessions/unified_XXXXXXXXX_XXXXXXXX HTTP/1.1" 404 Not Found

Disk check:
$ docker exec deeptutor find /app/data/multi-user -type f
(empty — directory never created)

### Additional Information

- DeepTutor Version: v1.3.10
- Operating System: Ubuntu 24.04
- Deployment: Docker (docker-compose.ghcr.yml) on Hetzner CPX22
- Admin account works perfectly — chat, TutorBot, Knowledge all functional
- Auth/login/user management UI all work correctly
- Root cause traced to get_current_path_service() in multi_user/paths.py — get_current_user_or_none() returns None for non-admin users during session requests, falling back to single-user path service. The multi-user/<uid>/ workspace is never initialized.

truffle-dev:
### Description

When `AUTH_ENABLED=true`, every non-admin HTTP request hits 404 because the user `ContextVar` set inside `require_auth` is invisible by the time the endpoint resolves the per-user path service.

`require_auth` is declared as a sync `def`. FastAPI dispatches sync deps via `anyio.to_thread.run_sync`, which captures the request context with `copy_context()` and runs the function in a worker thread under that copy. Mutations made inside the thread, including `ContextVar.set`, live on the copy and are discarded when the thread returns. The endpoint then runs in the original event-loop context with `_current_user` unset, so `get_current_path_service()` falls back to `PathService.get_instance()`, which points at `data/user/` — the admin workspace. Admin requests appear to work because the fallback coincides with their intended workspace; user requests resolve to the admin DB and 404 on every session lookup.

The user-context middleware that #474 introduced for exactly this case was removed in 8bae3ca on the assumption that `require_auth` was sufficient. With `require_auth` running in a threadpool, it was not.

Declaring `require_auth` (and `require_admin`, for symmetry) as `async def` keeps the dependency on the event loop; the `ContextVar` mutation now propagates to the endpoint and per-user path resolution works.

### Related Issues
- Closes #481

### Module(s) Affected
- [x] `api`
- [x] `tests`

### Checklist

### Additional Notes

Regression coverage in `tests/api/test_auth_contextvar.py` pins two invariants:

1. `require_auth` and `require_admin` are declared `async`.
2. With `AUTH_ENABLED=true` and a valid token, the user `ContextVar` set inside `require_auth` is visible from inside the endpoint, so `get_path_service()` resolves to the per-user `multi-user/<uid>/` workspace.

Verified locally with `pytest tests/api/test_auth_contextvar.py` — all four cases pass on the fix and fail on the prior sync `def`.

### DIFF
--- a/deeptutor/api/routers/auth.py
+++ b/deeptutor/api/routers/auth.py
@@ -157,7 +157,7 @@ def _extract_token(authorization: str | None, dt_token: str | None) -> str | Non
 # ---------------------------------------------------------------------------
 
 
-def require_auth(
+async def require_auth(
     authorization: str | None = Header(default=None, alias="Authorization"),
     dt_token: str | None = Cookie(default=None),
 ) -> TokenPayload | None:
@@ -168,11 +168,20 @@ def require_auth(
       - Authorization: Bearer <token> header
       - dt_token cookie
 
-    Works on both HTTP and WebSocket routes — ``Header`` and ``Cookie`` are
-    WS-compatible, while ``HTTPBearer`` (which we used to use here) is not.
+    ``Header`` and ``Cookie`` are kept here in place of ``HTTPBearer`` so the
+    function stays usable from WebSocket call sites that don't go through
+    FastAPI's standard HTTP request lifecycle.
 
     Returns the authenticated TokenPayload, or None if auth is disabled.
     Raises HTTP 401 if auth is enabled but the token is missing or invalid.
+
+    Declared ``async def`` so the ``set_current_user`` call runs in the same
+    asyncio context as the endpoint. A sync dependency is dispatched via
+    ``anyio.to_thread.run_sync``, which executes the function in a worker
+    thread under a *copy* of the request context; any ``ContextVar.set``
+    inside that thread is discarded when the thread returns, leaving the
+    endpoint to read the unset default. That regression was the root cause
+    of #481.
     """
     if not AUTH_ENABLED:
         from deeptutor.multi_user.context import set_current_user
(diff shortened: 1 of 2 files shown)


pancacake:
Thanks for your contribution!

truffle-dev:
Thanks for the quick review and merge.

PR #490, opened 2026-05-17
fix(visualize): unblock Gemini 2.5+ and harden Visualize pipeline

### LINKED ISSUE #489: [Bug]: Visualize (and other capabilities) silently truncate on Gemini 2.5 / 3.x due to default thinking tokens
## Summary

Gemini 2.5 / 3.x models ship with thinking enabled by default. When DeepTutor sends `max_tokens=4096` (the Visualize default) or even larger budgets, Gemini spends most of that budget on internal reasoning tokens and the model returns truncated output with `finish_reason=length` — often before any visible content has streamed. Pipelines that need structured output (Visualize codegen + review, Deep Solve writing, etc.) appear to half-produce results and either crash downstream parsers or render unusable artifacts.

## Repro

1. Configure DeepTutor with `LLM_BINDING=gemini`, `LLM_MODEL=gemini-2.5-flash` (any Gemini 2.5/3.x flash-tier model reproduces).
2. In the chat UI, switch the mode to **Visualize** and pick any non-trivial prompt — e.g.
   > Build an interactive long-division tutor for Year 7 that teaches the algorithm one digit at a time. Use the problem 7852 / 6. Walk me through divide / multiply / subtract / bring down for each digit with input boxes and a Check button.
3. Observe: stream halts mid-output (HTML cut off mid-CSS or mid-script, SVG cut off mid-tag). Codegen returns ~300-2000 chars and then stops.

Reproducible against the Gemini OpenAI-compat endpoint independently:

```bash
curl -s 'https://generativelanguage.googleapis.com/v1beta/openai/chat/completions' \
  -H \"Authorization: Bearer \$KEY\" -H 'Content-Type: application/json' \
  -d '{\"model\":\"gemini-2.5-flash\",\"stream\":false,\"max_tokens\":4096,
       \"messages\":[{\"role\":\"user\",
       \"content\":\"Write a 50-line SVG of cookies, just the SVG.\"}]}' \
  | jq '{finish:.choices[0].finish_reason, len:(.choices[0].message.content|length), usage}'
```

Result with thinking enabled (default):
\`\`\`
{ \"finish\": \"length\", \"len\": 360, \"usage\": {\"prompt_tokens\":24, \"completion_tokens\":164, \"total_tokens\":4116} }
\`\`\`

Result with `reasoning_effort: \"none\"` added to the request:
\`\`\`
{ \"finish\": \"stop\", \"len\": 2594, \"usage\": {\"prompt_tokens\":24, \"completion_tokens\":1426, \"total_tokens\":1450} }
\`\`\`

`total_tokens - prompt - completion` in the first call (~3900 tokens) is exactly the missing reasoning-token budget that ate the response.

## Why DeepTutor doesn't currently mitigate this

`OpenAICompatProvider._build_kwargs` (the live path) only auto-injects `reasoning_effort=\"high\"` when a model matches `spec.reasoning_model_patterns`. The Gemini `ProviderSpec` doesn't set those patterns. There's no default-down behavior for thinking-by-default models, so they silently consume the budget.

Also contributing (already filed as related smaller bugs but worth noting):

- The Visualize capability has no entry in `agents.yaml` / `DEFAULT_AGENTS_SETTINGS` / `loader.py:section_map`, so it silently uses the 4096-token default instead of a higher budget appropriate for HTML pages.
- `ReviewAgent.process` raises a JSONDecodeError that propagates up and kills the whole Visualize turn when the model returns prose instead of strict JSON (downstream consequence of the truncation above).

## Proposed fix

Disable thinking by default for Gemini 2.5/3.x in the three execution paths (`openai_compat_provider`, `executors`, `cloud_provider`) — caller can still opt in via explicit `reasoning_effort`. Plus the three smaller adjacent fixes (agents.yaml entry, review-stage graceful fallback, codegen tag-trim).

PR coming.

## Environment

- DeepTutor: current `dev` branch (cade789)
- LLM_BINDING=gemini, LLM_MODEL=gemini-2.5-flash
- Setup Tour install on macOS

skinred78:
Fixes #489.

## Summary

Six bugs that combined to break the Visualize capability on Gemini 2.5 Flash (and similar thinking-by-default models). Each is independently useful, but the user-visible symptom — Visualize output randomly truncated at ~370 chars — needs all of them.

### 1. Gemini 2.5/3.x reasoning-tokens default (root cause)

Gemini 2.5+ models burn most of `max_tokens` on internal "thinking" tokens by default. With `max_tokens=4096`, ~3900 tokens went to reasoning and ~160 came out as visible content, causing `finish_reason=length` on every multi-step pipeline (Visualize codegen + review, Deep Solve writing, etc.). See #489 for the reproducer with `curl` directly against Gemini's OpenAI-compat endpoint.

Default `reasoning_effort="none"` for Gemini 2.5/3.x when the caller doesn't specify, in all three execution paths:
- `provider_core/openai_compat_provider.py:_build_kwargs` (live path)
- `executors.py:sdk_complete` / `sdk_stream` (SDK path)
- `cloud_provider.py:_openai_complete` / `_openai_stream` (aiohttp fallback)

Callers that *want* thinking can still opt in via explicit `reasoning_effort`.

### 2. `visualize` capability had no `agents.yaml` entry

`get_agent_params("visualize")` silently fell through to the 4096-token default because there was no entry in `DEFAULT_AGENTS_SETTINGS` or `loader.py:section_map`. Added both, with a 16384-token budget appropriate for full HTML pages.

### 3. Review stage crashed hard on JSON parse failure

`ReviewAgent.process` does `ReviewResult.model_validate(extract_json_object(response))`. When the model returned prose instead of JSON (common with large SVGs that don't escape cleanly into a JSON string field), the parse raised and killed the entire Visualize turn — the user saw the streamed SVG draft and then a traceback. Wrapped `pipeline.run_review()` in try/except so a parse failure falls back to the unreviewed draft and the user still gets a rendered result.

### 4. Codegen output not trimmed to the root tag

Models often wrap SVG/HTML in prose ("Here you go: <svg>…</svg> Enjoy!") or emit a closing code fence on the same line as `</html>`, which `extract_code_block`'s regex (requiring a leading `\n` before the fence) doesn't strip. Added defensive root-tag trimming for `render_type=="svg"` and `render_type=="html"` so the renderer always sees a clean payload.

## Test plan

Verified end-to-end on Gemini 2.5 Flash via the CLI and via headless Playwright:
- Long-division HTML test prompt: 22 KB self-contained interactive page (was: 0 bytes / crash on `dev`).
- All step prompts read correctly (e.g. "How many whole times does 6 go into 7?" for the first digit).
- Walking the full 7852 ÷ 6 algorithm step by step terminates with `Quotient: 1308, Remainder: 4`.
- Wrong-answer retry preserves prior progress; 3 wrong attempts reveals the answer and auto-advances (graceful-fallback path exercised).
- Simple SVG (e.g. "draw 8 cookies") still renders unchanged — no regression on the happy path.
- The defensive root-tag trim is a no-op when codegen already returns a clean tag.

- [ ] Maintainer: please run `pre-commit run --all-files` and CI as a sanity check; my local environment doesn't have all the lint deps configured.

### DIFF
--- a/deeptutor/agents/visualize/agents/code_generator_agent.py
+++ b/deeptutor/agents/visualize/agents/code_generator_agent.py
@@ -84,4 +84,25 @@ class CodeGeneratorAgent(BaseAgent):
             if lowered.startswith("<!doctype") or lowered.startswith("<html"):
                 return stripped
 
+        # The renderers expect their payload to start cleanly with the right
+        # root tag. Models often wrap output with prose ("Here you go: <svg>…")
+        # or emit a code fence on the same line as the closing tag, which
+        # `extract_code_block`'s regex (requiring a leading \n before ```) does
+        # not strip. Trim to the outermost root tags as a defensive net.
+        if extracted:
+            if analysis.render_type == "svg":
+                lower = extracted.lower()
+                start = lower.find("<svg")
+                end = lower.rfind("</svg>")
+                if start != -1 and end != -1 and end > start:
+                    extracted = extracted[start : end + len("</svg>")]
+            elif analysis.render_type == "html":
+                lower = extracted.lower()
+                start = lower.find("<!doctype")
+                if start == -1:
+                    start = lower.find("<html")
+                end = lower.rfind("</html>")
+                if start != -1 and end != -1 and end > start:
+                    extracted = extracted[start : end + len("</html>")]
+
         return extracted
--- a/deeptutor/capabilities/visualize.py
+++ b/deeptutor/capabilities/visualize.py
@@ -8,6 +8,7 @@ Produces SVG or Chart.js code from user requests and conversation context.
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 from deeptutor.capabilities.request_contracts import get_capability_request_schema
@@ -15,6 +16,8 @@ from deeptutor.core.capability_protocol import BaseCapability, CapabilityManifes
 from deeptutor.core.context import UnifiedContext
 from deeptutor.core.stream_bus import StreamBus
 from deeptutor.core.trace import merge_trace_metadata
+
+logger = logging.getLogger(__name__)
 
 
 class VisualizeCapability(BaseCapability):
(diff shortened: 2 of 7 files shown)


pancacake:
Thanks for your contribution!

PR #509, opened 2026-05-23
fix: prevent Document nodes from bypassing chunking pipeline

washi4:
## Summary

Fixes a bug where all documents bypass `SentenceSplitter` during KB indexing, causing full-document text (1M+ chars) to be sent to the embedding API and triggering HTTP 400 input-length errors.

## Root Cause

LlamaIndex's `Document` class inherits from `BaseNode`. The previous filtering logic:

```python
text_documents = [doc for doc in documents if not isinstance(doc, BaseNode)]
```

classified **every** Document as a pre-embedded node, skipping the chunking pipeline entirely.

## Fix

Replace the naive `isinstance(doc, BaseNode)` check with `_has_precomputed_embedding()` that:
1. Always treats `Document` instances as needing chunking
2. Only skips nodes that are non-Document `BaseNode` subclasses **and** already carry an embedding vector (e.g. `ImageNode` from multimodal loaders)

## Testing

Verified end-to-end: KB creation with a 1M+ char PDF now correctly chunks into 512-token segments and embeds successfully via DashScope `text-embedding-v3`.

### DIFF
--- a/deeptutor/services/rag/pipelines/llamaindex/ingestion.py
+++ b/deeptutor/services/rag/pipelines/llamaindex/ingestion.py
@@ -12,7 +12,7 @@ from typing import Any
 from llama_index.core import Settings, VectorStoreIndex
 from llama_index.core.ingestion import IngestionPipeline
 from llama_index.core.node_parser import SentenceSplitter
-from llama_index.core.schema import BaseNode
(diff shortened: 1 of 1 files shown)


pancacake:
Thanks for your contribution!

PR #522, opened 2026-05-26
feat: support zip upload with error-state recovery and recursive file scanning

xiongjnu:
## Summary

Three related improvements to the knowledge base file upload pipeline:

1. **ZIP archive upload support** — Users can now upload .zip files containing knowledge base documents. The archive is extracted server-side, files are validated against allowed extensions, and individual files are processed normally. Path traversal attempts (../, absolute paths) in zip entries are rejected with a 400 error.

2. **Error-state recovery** — When uploading files to a knowledge base that is in an error state, the upload endpoint now falls back to a full KnowledgeBaseInitializer.process_documents() run instead of failing on the incremental DocumentAdder path.

3. **Recursive file scanning** — FileTypeRouter.collect_supported_files() calls are updated to use recursive=True, so files in nested subdirectories inside raw/ are discovered during indexing, metadata collection, and reindexing.

## Changes

| File | Change |
|------|--------|
| deeptutor/api/routers/knowledge.py | Zip extraction in _save_uploaded_files(); error-state fallback in run_upload_processing_task(); .zip added to allowed extensions; recursive=True for file collection |
| deeptutor/knowledge/initializer.py | recursive=True for 3 collect_supported_files() calls |
| web/lib/knowledge-helpers.ts | kbIsUploadable now excludes initializing/processing states instead of requiring === ready, allowing upload to error-state KBs |

## Test plan

- [ ] Upload a .zip containing valid PDF/markdown files to a KB -> files are extracted and indexed
- [ ] Upload a .zip with path traversal entries -> 400 error, safe extraction blocked
- [ ] Upload a .zip containing an invalid .exe file -> silently skipped (not in allowed extensions)
- [ ] Upload files to a KB in error state -> auto-recovery via full initialization
- [ ] Upload files to a KB with nested subdirectories in raw/ -> all files found during indexing

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,51 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+# deeptutor
+
+HKU DeepTutor（GitHub 21k ⭐，Apache-2.0）— 开源 AI 学习辅导系统。提供 Chat、Deep Solve、Quiz Generation、Math Animator 等 6 大学习模式。
+
+## 技术栈
+
+- 后端：Python + FastAPI
+- 前端：Next.js (web/)
+- AI：DeepSeek API（已配置）
+- 向量：LlamaIndex RAG + bge-m3 embedding
+- 数据库：SQLite + 文件系统
+
+## 目录结构
+
+```
+deeptutor/
+├── deeptutor/           核心库（agents, tools, services, knowledge...）
+├── deeptutor_cli/       CLI 接口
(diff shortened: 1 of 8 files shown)


pancacake:
Thanks for your contribution! The zip-upload direction here is genuinely useful. We've taken it forward with a dedicated, safer implementation — per-entry extraction with Zip-Slip, zip-bomb (size/count/ratio) and extension-whitelist guards rather than a blanket extractall — plus the error-state recovery idea. Closing this PR in favour of that implementation, but really appreciate you surfacing the need and the approach.

PR #524, opened 2026-05-26
feat: truncate oversized event payloads in session response

xiongjnu:
RAG tool results can return KB document content exceeding 1MB per event, causing the frontend TraceSurface to run out of memory when loading large sessions. Truncate tool_result and observation payloads at 1MB in the `get_session` endpoint, marking truncated events with `_truncated: true`.

The LLM context builder reads from a separate content-only store and is unaffected by this truncation.

## Test plan
- [ ] Create a session with RAG tool results exceeding 1MB → events are truncated, `_truncated: true` is set
- [ ] Small sessions (<1MB events) → no truncation occurs
- [ ] LLM context building → unaffected (uses separate data path)

### DIFF
--- a/deeptutor/api/routers/sessions.py
+++ b/deeptutor/api/routers/sessions.py
@@ -4,6 +4,7 @@ Unified session history API.
 
 from __future__ import annotations
 
+import json
 import logging
 
 from fastapi import APIRouter, HTTPException, Query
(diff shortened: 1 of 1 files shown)


pancacake:
Thanks for your contribution!

PR #525, opened 2026-05-26
feat: embed local Markdown image references as data URIs during indexing

xiongjnu:
When a Markdown document references local images (e.g. ``), resolve them relative to the markdown file and inline as base64 data URIs before passing the document to the embedding pipeline. This ensures images packaged in a zip upload survive the extraction and indexing step without broken links.

External URLs and data URIs are left untouched. Images exceeding 500 KB are skipped to keep the index size reasonable.

This is a companion to PR #522 (zip upload support) — without image inlining, markdown files extracted from a zip archive would have broken local image references.

## Test plan
- [ ] Index a .md file with `` referencing a local image → image is inlined as data URI in the document text
- [ ] Index a .md file with external URLs `` → left untouched
- [ ] Index a .md file referencing a non-existent image → kept as-is
- [ ] Index a .md file referencing an image >500 KB → skipped, kept as-is

### DIFF
--- a/deeptutor/services/rag/pipelines/llamaindex/document_loader.py
+++ b/deeptutor/services/rag/pipelines/llamaindex/document_loader.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import logging
 import mimetypes
+import re
 from pathlib import Path
 from typing import Any, Iterable
 
@@ -28,6 +29,11 @@ IMAGE_DESCRIPTION_PROMPT = (
     "and cite it later. Include visible text/OCR if present, the main subject, "
     "and any educational or technical meaning. Keep the answer under 180 words."
 )
+
+# Match Markdown image references: ![alt](path) or ![](path)
+_MD_IMAGE_RE = re.compile(r'!\[([^\]]*)\]\(([^)]+)\)')
+# Max size for embedding referenced images as data URIs
+_MAX_EMBED_IMAGE_SIZE = 500 * 1024  # 500 KB
 
 
 class LlamaIndexDocumentLoader:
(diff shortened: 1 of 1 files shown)


polaris-smart:
Thanks for this PR — the approach is clean and well-scoped. I tested the happy path in our DeepTutor deployment (LlamaIndex pipeline, real KBs): local `` references are correctly inlined as data URIs, and external URLs / data URIs are left untouched as described.

**One security concern before this can merge: unrestricted path resolution turns this into an arbitrary-file-read primitive.**

`_resolve_markdown_images` resolves `img_path` via `(md_file_path.parent / img_path).resolve()` with no constraint on where the resolved path lands. Both forms read files outside the document's directory:

```markdown
![](/app/data/user/settings/model_catalog.json)   ← absolute path wins the pathlib join
![](../user/settings/model_catalog.json)          ← traversal
```

Verified locally: a ~21 KB `model_catalog.json` (which holds provider API keys in a default deployment) gets base64-inlined into the document text as `data:application/json;base64,...`, then chunked and indexed. Any user who can upload a markdown file to their own KB can later retrieve that chunk from their own retrieval results, decode it, and read the keys. `mimetypes.guess_type` returning `None` doesn't help — the fallback to `application/octet-stream` inlines the file anyway.

Suggested hardening (small):

1. Skip absolute paths (`Path(img_path).is_absolute()`).
2. After `.resolve()`, require the result to stay inside the markdown file's parent directory (e.g. `resolved.is_relative_to(md_file_path.parent.resolve())`) — otherwise return the original match unchanged.
3. Restrict to image extensions (png/jpg/jpeg/gif/webp). With (1)+(2) this is defense-in-depth, but it makes the intent explicit and protects against future refactors.

Minor non-blocking note: `[^)]+` captures `path "title"` wholesale for the `alt` form, so titled images silently skip inlining. Fine to leave, but worth a code comment.

Happy to re-run our test matrix (inline / external / missing / oversize) once the hardening lands.

polaris-smart:
Update from our side: we've carried a hardened variant of this feature in our deployment fork ([commit 267aad10](https://github.com/HKUDS/DeepTutor/commit/267aad10)) and ran a 7-case matrix against it — inline / traversal / absolute / external / oversize / decode-all-no-leak / titled-image. All pass. If it's helpful, we're happy to push the hardened  as commits to this PR so the fix and the feature land together.

polaris-smart:
Update from our side: we've carried a hardened variant of this feature in our deployment fork (commit `267aad10` on our internal mirror) and ran a 7-case matrix against it — inline / traversal / absolute / external / oversize / decode-all-no-leak / titled-image. All pass. If it's helpful, we're happy to push the hardened resolver (absolute-path rejection + directory containment + image-extension allowlist) as commits to this PR so the fix and the feature land together. (Supersedes my previous comment, which lost a code span to shell quoting.)

PR #549, opened 2026-06-04
fix: keep NEXT_PUBLIC_AUTH_ENABLED out of the minifier constant-fold

IliaAvdeev:
### Description
With auth enabled, none of the auth UI shows up in the Docker image — no Admin
link, no Sign out, and the 401 → /login redirect never fires — even though the
backend auth works fine.

The problem is how the build-time flag is read. Both `lib/auth.ts` and
`lib/api.ts` do:

    const AUTH_ENABLED = process.env.NEXT_PUBLIC_AUTH_ENABLED === "true";

In the Docker build the value is injected as the literal
`__NEXT_PUBLIC_AUTH_ENABLED_PLACEHOLDER__`, and `start-frontend.sh` rewrites it
on container start. But since this is a `"<literal>" === "true"` comparison, the
minifier folds it down to `false` at build time and drops the placeholder
completely. By the time the startup script runs there's nothing left to
substitute, so the client flag is stuck at `false`.

`NEXT_PUBLIC_API_BASE` avoids this because it's used as a plain string and never
compared against a literal, so its placeholder survives minification.

The fix reads the raw value and evaluates it at runtime, which keeps the
placeholder in the bundle:

    const AUTH_ENABLED = /^(1|true|yes|on)$/i.test(
      (process.env.NEXT_PUBLIC_AUTH_ENABLED ?? "").trim(),
    );

After this, `start-frontend.sh` can substitute the placeholder as intended and
the auth UI appears when auth is on.

How to reproduce on `dev`: build the image, enable auth (`auth.json`
`enabled: true` or `AUTH_ENABLED=true`), start the container, and grep the built
bundle — `AUTH_ENABLED` is baked as `false` and the placeholder is gone. With
this change the placeholder is preserved and rewritten to `true` at startup.

---

Notes:
- Ran `pre-commit run --all-files`; my changes pass. Two unrelated failures
  already exist on `dev` and are left untouched as out of scope: a duplicate key
  in `web/locales/{en,zh}/app.json` (check-json), and Unix-only `os.getpgid` /
  `killpg` in `runtime/launcher.py` that mypy flags on Windows.
- Added unit tests (`web/tests/auth-enabled.test.ts`, node:test) for the flag
  parsing — accepted truthy values, falsy/empty/undefined, and that an
  unsubstituted Docker placeholder reads as disabled. After extracting the
  shared `parseAuthEnabled` helper, the placeholder still survives minification
  (verified with a production build), so the runtime rewrite is unaffected.

### Module(s) Affected
- [ ] `agents`
- [ ] `api`
- [ ] `config`
- [ ] `core`
- [ ] `knowledge`
- [ ] `logging`
- [ ] `services`
- [ ] `tools`
- [ ] `utils`
- [x] `web` (Frontend)
- [ ] `docs` (Documentation)
- [ ] `scripts`
- [x] `tests`
- [ ] Other: `...`

### Checklist
- [ ] I have added relevant tests for my changes.

### DIFF
--- a/web/lib/api.ts
+++ b/web/lib/api.ts
@@ -131,12 +131,22 @@ export function wsUrl(path: string): string {
   return `${normalizedBase}${normalizedPath}`;
 }
 
-// Evaluate the build-time flag at runtime (not a ``=== "true"`` compare) so the
-// Docker ``__NEXT_PUBLIC_AUTH_ENABLED_PLACEHOLDER__`` token survives minification
-// and ``start-frontend.sh`` can rewrite it on startup. See lib/auth.ts.
-const AUTH_ENABLED = /^(1|true|yes|on)$/i.test(
-  (process.env.NEXT_PUBLIC_AUTH_ENABLED ?? "").trim(),
-);
+/**
+ * Parse a ``NEXT_PUBLIC_AUTH_ENABLED``-style flag at runtime.
+ *
+ * The value is inlined at build time and, in the Docker image, baked as the
+ * ``__NEXT_PUBLIC_AUTH_ENABLED_PLACEHOLDER__`` token that ``start-frontend.sh``
+ * rewrites on container start. Evaluating it with a runtime regex — instead of
+ * a ``=== "true"`` literal comparison the minifier would constant-fold — keeps
+ * the placeholder in the bundle so it survives minification (the same approach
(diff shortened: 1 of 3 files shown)


IliaAvdeev:
Added tests

PR #588, opened 2026-06-23
fix: render LaTeX in quiz options

### LINKED ISSUE #587: Options do not render LaTeX formulas in quiz cards
## Summary

Multiple-choice option text in generated quiz cards can display raw LaTeX source instead of rendered math.

## Current behavior

In the unified chat quiz card, the question stem renders LaTeX correctly, but answer options show escaped/source text such as raw inline math delimiters and commands. Examples include \(1\), \(x\ne -3,0,3\), and \(\dfrac{x-3}{x}\).

This is visible in the option rows (A/B/C/D), while the same card body renders math formulas correctly.

## Expected behavior

LaTeX contained in option labels should be rendered the same way as LaTeX in the question body, so students see formatted math rather than source markup.

## Reproduction context

1. Start the web app and open the unified chat UI.
2. Generate or view a multiple-choice math question whose answer options contain inline LaTeX.
3. Observe that option rows display raw LaTeX delimiters and commands instead of rendered math.

## Impact

This makes math answer choices hard to read and can obscure the intended answer, especially for expressions with fractions, inequalities, or excluded values.

ZhouJ-sh:
## Summary

- render multiple-choice option labels with the existing MarkdownRenderer math pipeline
- add a regression test that prevents option labels from falling back to raw text rendering

Fixes #587

## Checks

- npm run test:node
- pre-commit run --all-files

### DIFF
--- a/web/components/quiz/QuizViewer.tsx
+++ b/web/components/quiz/QuizViewer.tsx
@@ -1038,7 +1038,13 @@ export default function QuizViewer({
                       key
                     )}
                   </span>
-                  <span className="leading-relaxed">{text}</span>
+                  <span className="min-w-0 leading-relaxed">
+                    <MarkdownRenderer
+                      content={text}
+                      variant="compact"
+                      enableMath
+                    />
+                  </span>
                 </button>
               );
             })}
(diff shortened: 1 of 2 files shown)


pancacake:
Merged into `dev` with a small fixup: the option label is wrapped in a `<div>` instead of a `<span>` so the block-level `MarkdownRenderer` output (`<div><p>`) stays valid inside the option `<button>` (avoids React's `validateDOMNesting` warning) and matches the four sibling `MarkdownRenderer` call sites in the file. The serif/prose option text is kept intentionally — it matches the question stem. Thanks!

PR #600, opened 2026-06-27
Adaptive diagnostic: prefer nearest difficulty on fallback

anideebee7:
## Summary
- Fixes the "adaptive but behaves like a fixed rotation" bug in the diagnostic engine.
- When the target difficulty has no question in the pack, `_pick_diagnostic_question_relaxed` previously iterated `DIFFICULTY_LEVELS` in fixed easy→hard order, so a correct MEDIUM answer (which should step UP) could be served an EASY question — masking real-time adaptation.
- Adds `_difficulty_fallback_order(target, prefer_harder)`: tries the target, then the **nearest** available difficulty, using the previous answer's correctness as the tie-break when target is `medium`.
- `diagnostic_answer` now passes the `is_correct` signal into the picker so served difficulty tracks performance even with a sparse bank.

## Notes
- Additive/behavioral fix scoped to `deeptutor/api/routers/exams.py`; no API or schema change.
- Residual (separate, data not code): seeding a dense difficulty×subject bank so the fallback rarely triggers.

## Test plan
- [ ] Diagnostic: a correct answer steps difficulty up (or to nearest-higher when target absent); a wrong answer steps down.
- [ ] Sparse pack still returns a question (no dead end) and the served difficulty is the nearest to target.


anideebee7:
Opened against the wrong repo (fork parent). Re-opening on intelli-verse-x/DeepTutor.

PR #628, opened 2026-07-15
fix: resolve relative img_path in parsed content lists to absolute paths

### LINKED ISSUE #624: [Bug] local mineru+lightrag

### Describe the bug

When creating a knowledge base or adding a document to an existing knowledge base using mineru+lightrag, a problem arises when generating image descriptions. An error occurs when attempting to locate image files via the relative path `image/*.png`. If the path is manually corrected to an absolute path in the `*content*` files after mineru has parsed them, everything works fine.

All packages are the latest versions... (lightrag==1.4.16)

Thank you

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Related Module

Dashboard

### Configuration Used

_No response_

### Logs and screenshots

_No response_

### Additional Information

- DeepTutor Version:
- Operating System:
- Python Version:
- Node.js Version:
- Browser (if applicable):
- Related Issues:

xor-xe:
### Description
When indexing documents with MinerU + LightRAG, image description generation fails because content-list `img_path` entries are relative (`images/*.png`). They are relative to the content-list file inside DeepTutor's parse cache, but RAG-Anything resolves them against its own working directory, so the images are never found.

This PR anchors relative `img_path` values at the content dir in `cache.load_ir`, the shared loader all ParseService consumers use. The cached JSON stays relative on disk so cache dirs remain relocatable; absolute paths pass through unchanged.

### Related Issues
- Closes #624

### Module(s) Affected
- [ ] `agents`
- [ ] `api`
- [ ] `config`
- [ ] `core`
- [ ] `knowledge`
- [ ] `logging`
- [x] `services`
- [ ] `tools`
- [ ] `utils`
- [ ] `web` (Frontend)
- [ ] `docs` (Documentation)
- [ ] `scripts`
- [x] `tests`
- [ ] Other: `...`

### Checklist

### Additional Notes
Reproduced the failure path in a unit test: `load_ir` returned relative `img_path` values straight from the content-list JSON. Verified with `tests/services/parsing` and `tests/services/rag/test_lightrag_pipeline.py` (59 passed, 2 skipped) plus `ruff check` / `ruff format --check` on the touched files.

### DIFF
--- a/deeptutor/services/parsing/cache.py
+++ b/deeptutor/services/parsing/cache.py
@@ -123,6 +123,24 @@ def find_content_dir(workdir: Path) -> Path:
     return candidate_dirs[0] if candidate_dirs else workdir
 
 
+def _absolutize_img_paths(blocks: list[dict], content_dir: Path) -> list[dict]:
+    """Anchor relative ``img_path`` entries at ``content_dir``.
+
+    Engines (MinerU especially) emit ``img_path`` values like ``images/x.png``
+    relative to the content-list file. Consumers such as RAG-Anything resolve
+    them against their own working directory instead, so the images are never
+    found (issue #624). The cached JSON stays relative — paths are rewritten
+    only on load, keeping the cache dir relocatable.
+    """
+    for block in blocks:
+        if not isinstance(block, dict):
+            continue
+        img_path = block.get("img_path")
+        if img_path and not Path(img_path).is_absolute():
+            block["img_path"] = str(content_dir / img_path)
+    return blocks
+
+
 def load_ir(workdir: Path) -> tuple[str, Optional[list[dict]], Optional[Path]]:
     """Load ``(markdown, blocks, asset_dir)`` from a parsed/cached dir.
 
@@ -143,7 +161,7 @@ def load_ir(workdir: Path) -> tuple[str, Optional[list[dict]], Optional[Path]]:
         try:
             loaded = json.loads(json_files[0].read_text(encoding="utf-8"))
             if isinstance(loaded, list):
-                blocks = loaded
+                blocks = _absolutize_img_paths(loaded, content_dir)
         except Exception as exc:
             logger.warning("Failed to read content_list %s: %s", json_files[0], exc)
 
--- a/tests/services/parsing/test_cache.py
+++ b/tests/services/parsing/test_cache.py
@@ -55,6 +55,33 @@ def test_load_ir_reads_markdown_blocks_images(tmp_path: Path) -> None:
     assert asset_dir == workdir / "images"
 
 
+def test_load_ir_absolutizes_relative_img_paths(tmp_path: Path) -> None:
+    workdir = tmp_path / "wd"
+    workdir.mkdir()
+    (workdir / "doc.md").write_text("# title", encoding="utf-8")
+    (workdir / "doc_content_list.json").write_text(
+        json.dumps(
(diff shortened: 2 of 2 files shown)


rickgonzalez:
a scanned explainer from Lenzon - I hope it helps explain this PR https://www.lenzon.ai/viewer/cmrmdcjxr0005zmi33qnbpuf1?voice=google-chirp3

riju568:
The Core Problem
DeepTutor makes extensive use of RAG and document extraction engines (e.g., LlamaIndex, LightRAG) for feeding educational markdown, pdf and learning documents into its Knowledge Center. When users upload these documents, inline images and figures within the document frequently refer to the local relative file structure (../assets/diagram.png, ./images/formula.jpg, and the like).

Previously, when our backend read through those documents and assembled the parsed content lists that were then handed off to either the multi-agent solving loop or to the front-end interface, it would hand over those strings, img_path, with no modification as they were entered by the user.

This causes two major failures across DeepTutor's architecture:
1. Broken UI assets: Next.js frontend UI / Electron client resolve the relative assets string against their own domain, e.g. To look for “/assets/diagram.png” on http://127.0.0.1:3782. Due to dynamic routes / work directory the relative assets would resolve into an invalid URL likehttp://127.0.0.1:3782/assets/diagram.png and lead to a “Not Found” error / broken image in the chat workspace.
2. Context Fragmentation For Multimodal Agents In both loops of investigation (InvestigateAgent, NoteAgent), we use stable chunks of a document as the context. However, if the context path for the image is broken, any other process involving multimodal interaction or text retrieval does not gain access to the required image data.

xor-xe:
This PR is narrower in scope than what you're describing. It only fixes the backend indexing path: MinerU writes img_path values relative to the content-list file in the parse cache, and RAG-Anything resolves them against its own working directory, so the vision stage can't open the images during LightRAG ingestion (#624). The fix anchors those paths at load time in cache.load_ir.

Frontend asset resolution isn't touched here namely the web UI doesn't consume content-list img_path values. Also note InvestigateAgent/NoteAgent were removed from main in 23ca3024 (the agentic engine refactor), so that part refers to the pre-refactor architecture.

riju568:
This PR is for the Pull Request #628 of the HKUDS/DeepTutor project and the issue that is being addressed is as follows.

# The Issue

We found out that when indexing documents with MinerU + LightRAG to retrieve the image descriptions, it failed. The reason was that the img_path values in the parsed contents' lists are being stored relative to the content-list file path under cache/parse such that they look like images/*.png.

The paths are indeed relative to the content-list.json within the parse cache folder under cache, but when those documents are later processed by RAG-Anything, it tries to look them up using RAG-Anything's directory as reference. Because the paths don't match, the vision stage is not able to find and open them for LightRAG during the ingestion (which has been traced to Issue #624).

# The Fix

This PR fixes it by anchoring the imgpaths on load with cache.loadir, which is the common loader that ParseService consumes. In short, it will:

* Make the relative img_paths reference to the correct content-directory on load so that RAG-Anything is able to find and use them.
* Keep the img_paths in the cached JSON documents relative, so that the parse cache directory could be arbitrarily placed without losing access.

Important note: Absolute paths will not be altered at all. Also, this is only for the backend indexing path, not for any web asset loading (e.g., what the web UI serves for itself) in other parts of the system, as brought up by other collaborators in the comments.

PR #635, opened 2026-07-16
fix(knowledge): update KB status to processing on file upload and sync-folder

Marnie0415:
### Description

When files are uploaded to an existing knowledge base via `POST /{kb_name}/upload`, or when a linked folder is synced via `POST /{kb_name}/sync-folder/{folder_id}`, the KB status remains "ready" until the background indexing task starts executing. During this window — which can be several seconds under load — the user sees the KB as fully indexed and ready for search, but the newly uploaded files are not yet searchable.

This creates a confusing UX: the user uploads files, immediately searches, gets no results, and assumes the upload failed.

The `create_knowledge_base` endpoint already handles this correctly by calling `manager.update_kb_status(name=name, status="initializing", ...)` before starting the background task. The `upload_files` and `sync_folder` endpoints omit this step.

### What this changes

Added `manager.update_kb_status()` calls in both `upload_files` and `sync_folder` that set the KB status to "processing" before the background indexing task is dispatched. The ProgressTracker inside the background task will later update the status to "ready" or "error" when indexing completes.

### User impact

After uploading files or syncing a linked folder, the KB status immediately reflects that indexing is in progress, matching the behavior of `create_knowledge_base`. Users no longer see a stale "ready" status during the indexing window.

### Files affected

- `deeptutor/api/routers/knowledge.py` — added status updates in `upload_files` and `sync_folder`

### Testing

The existing test suite covers `upload_files`, `sync_folder`, and `create_knowledge_base`. The status updates use the same `manager.update_kb_status()` call that `create_knowledge_base` already uses, so the behavior is consistent.

### Related Issues
- None

### Module(s) Affected
- [x] `api`
- [x] `knowledge`


Marnie0415:
@pancacake Hi, I've updated this PR based on review feedback:

1. **Added sync_folder status update** — the same race condition existed in the sync-folder endpoint, now fixed with the same pattern.
2. **PR description updated** — reflects both upload_files and sync_folder fixes.

Both changes use the existing manager.update_kb_status() call that create_knowledge_base already uses. The ProgressTracker in the background task will overwrite this status once indexing starts, so there's no conflict.

Ready for another look when you get a chance. Thanks!

PR #636, opened 2026-07-16
fix(book): reset page status on compilation failure (worker + inline)

Marnie0415:
### Description

When a Book page fails to compile, the page status is left stuck in `GENERATING` permanently. The user sees the page spinning forever with no way to recover.

This affects two code paths:

1. **Background worker** (`_worker_loop`): pages queued for background compilation. If the compiler throws (LLM timeout, network error, API rate limit), the `except` block logs the error but never resets the page status.

2. **Inline compile** (`compile_page` method): pages compiled on-demand when a user opens them. The same lack of exception handling means the page gets stuck in `GENERATING` if the compiler fails.

The root cause in both cases: `compiler.py` sets `page.status = PageStatus.GENERATING` before LLM calls, and `_finalize_page_status()` is only reached on success.

Additionally, the frontend `handleSelectPage` (page.tsx:342) skips GENERATING pages — clicking a stuck page does nothing. Only the "Force regenerate" button can recover, but users may not know it exists.

### What this changes

1. In `_worker_loop`: when background compilation fails, page status is reset to `ERROR` with the error message.
2. In `compile_page`: when inline compilation fails, page status is reset to `ERROR` with the error message, then the exception is re-raised so the caller still sees the error.

Both paths use the same `PageStatus.ERROR` assignment pattern already used in `compiler.py:351` and `compiler.py:361`.

### User impact

When compilation fails, the page shows an error state with a descriptive message instead of spinning forever. Users can recover via the "Force regenerate" button in the UI.

### Files affected

- `deeptutor/book/engine.py` — added error handling in `_worker_loop` and `compile_page`

### Testing

The existing Book Engine test suite covers both the worker loop and inline compilation paths. The fixes add exception handling in existing error paths, using the same `PageStatus.ERROR` pattern already established in the codebase.

### Related Issues
- None

### Module(s) Affected
- [x] `book`

### DIFF
--- a/deeptutor/book/engine.py
+++ b/deeptutor/book/engine.py
@@ -749,14 +749,7 @@ class BookEngine:
                     language=book.language,
                 )
         except Exception as exc:
-            # compiler sets page.status = GENERATING before LLM calls; if it
-            # throws, the status is left there permanently.  Reset to ERROR so
-            # the user sees the failure and can force-regenerate.
-            if page.status == PageStatus.GENERATING:
-                page.status = PageStatus.ERROR
-                page.error = f"Compilation failed: {exc}"
-                page.updated_at = time.time()
-                self.storage.save_page(page)
+            self._mark_page_error(page, exc, prefix="Compilation failed")
             raise
 
         # Refresh KB fingerprints once we successfully ship a READY page.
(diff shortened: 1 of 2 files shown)


Marnie0415:
@pancacake Hi, I've updated this PR based on review feedback:

1. **Added `compile_page` inline path fix** — the user-triggered compilation path had the same stuck-GENERATING bug. Now wrapped in try/except with status reset to ERROR.
2. **Fixed PR description** — removed the incorrect `resume_book` reference. Corrected to explain the actual recovery path: the "Force regenerate" button in the UI.
3. **Added frontend context** — noted that `handleSelectPage` skips GENERATING pages, so users can't click into a stuck page.

Both `_worker_loop` and `compile_page` now use the same `PageStatus.ERROR` assignment pattern already established in `compiler.py`. The `raise` in `compile_page` preserves the original exception so callers still see the failure.

Ready for another look. Thanks!

jlr2193224-code (approved):

pancacake:
Thanks for the contribution! Will take a look very soon. Thanks!

PR #637, opened 2026-07-17
Fix recency-weighting variable swap in compute_mastery

### LINKED ISSUE #618: compute_mastery recency weighting is disabled by a variable-swap in the zip unpack
In `deeptutor/learning/mastery.py`, `compute_mastery` is documented as *"recency-weighted accuracy"* (newer attempts count more, so recovery after early mistakes is rewarded). But the scoring line binds the loop variables in the wrong order:

```python
score = sum(w * (1.0 if c else 0.0) for w, c in zip(recent, weights, strict=True)) / sum(weights)
```

`recent` is the correctness list and `weights` is the recency weights, so `zip(recent, weights)` yields `(correctness, weight)` pairs — but they're unpacked as `w, c`, binding **`w` to the outcome** and **`c` to the weight**. Since a weight is always truthy, `(1.0 if c else 0.0)` is always `1.0`, so each term reduces to the raw correctness bit. The result:

- the **numerator becomes a plain correct-count** (recency weights only affect the denominator), so
- mastery is **order-independent** — a recovering history scores the same as a declining one.

**Minimal repro:**
```python
>>> compute_mastery([False, True, True])   # miss then two hits (recovering)
0.714...
>>> compute_mastery([True, True, False])   # two hits then a recent miss
0.714...   # identical — recency has no effect
```

**Expected:** `[F,T,T]` should score **higher** than `[T,T,F]` (recent attempts weigh more).

**Fix** — swap the unpack to match `zip`'s order:
```python
score = sum(w * (1.0 if c else 0.0) for c, w in zip(recent, weights, strict=True)) / sum(weights)
```

After the fix: `[F,T,T] = 0.696` vs `[T,T,F] = 0.643`; the confidence caps are unaffected.

smz202000:
Fixes #618. Zip(recent, weights) produces (correctness, weight) pairs, but the generator unpacked them as (w, c), assigning correctness to w and weight to c. Since weights are always truthy, the numerator was always a plain correct-count, disabling recency weighting. Fix: swap unpacking to (c, w).

### DIFF
--- a/deeptutor/learning/mastery.py
+++ b/deeptutor/learning/mastery.py
@@ -31,7 +31,7 @@ def compute_mastery(correctness: list[bool]) -> float:
         return 0.0
     recent = correctness[-len(_RECENCY_WEIGHTS) :]
     weights = _RECENCY_WEIGHTS[-len(recent) :]
-    score = sum(w * (1.0 if c else 0.0) for w, c in zip(recent, weights, strict=True)) / sum(
+    score = sum(w * (1.0 if c else 0.0) for c, w in zip(recent, weights, strict=True)) / sum(
         weights
     )
     return min(score, _CONFIDENCE_CAP.get(len(recent), 1.0))


paslaycorp:
Im so green to github i dont know ow where to find out where my url for public access is.
Is this in reference to my code?

PR #644, opened 2026-07-19
fix(partners): avoid hang in split_message when max_len <= 0

santhreal:
When max_len is 0 or negative, the cut pointer never advances and split_message loops forever. Return the content unsplit so Discord/Telegram/Weixin outbound paths cannot wedge.

Repro: split_message("hello world", max_len=0).


pancacake:
Thanks for your contribution!

PR #645, opened 2026-07-19
fix(memory): parse line-edit JSON with raw_decode

santhreal:
parse_edits_payload is a tolerant LLM JSON parser, but _extract_json used a greedy rfind("}") slice. Trailing prose that itself contains braces made the slice invalid, so valid edits were silently dropped. Decode the first JSON value with raw_decode (same approach as math_animator).

Repro: `{"edits":[...]} trailing {brace}` previously returned [].


pancacake:
Thanks for your contribution!

PR #646, opened 2026-07-19
fix(memory): parse update facts JSON with raw_decode

santhreal:
_parse_facts is a tolerant LLM JSON parser, but _extract_json_object used a greedy rfind("}") slice. Trailing prose that itself contains braces made the slice invalid, so valid facts were silently dropped. Decode the first JSON object with raw_decode (same approach as line-edit parsing).

Repro: `{"facts":[...]} trailing {note}` previously returned [].


pancacake:
Thanks for your contribution!

PR #658, opened 2026-07-21
fix(book): stream long-running operations over WebSocket

### LINKED ISSUE #607: [Bug]: confirm-proposal times out client-side (~35-47s) but backend finishes the spine build anyway

### Describe the bug

`POST /api/v1/book/books/confirm-proposal` ("Confirm proposal & build spine" button) reliably fails client-side with a proxy error after ~35-47s, even though the backend keeps working and successfully finishes the spine generation shortly after.

Frontend/Next.js log:
```
Failed to proxy http://localhost:8001/api/v1/book/books/confirm-proposal Error: socket hang up
    at ignore-listed frames {
  code: 'ECONNRESET'
}
```

Backend log: no error, no traceback, no `confirm_proposal failed` line at all (see `book.py` — `confirm_proposal` route logs `logger.error(f"confirm_proposal failed: {exc}", exc_info=True)` on exception, and this never appears). The backend is still mid-request when the connection drops.

I confirmed this by checking `GET /api/v1/book/books` on the backend directly after the failures: both affected books ended up in `spine_ready` / `compiling` state, with `created_at` timestamps matching the exact moment of the failed proxy calls. So the backend genuinely finishes the (multi-LLM-call) spine generation — the client/proxy just gives up first and the UI has no way to know it eventually succeeded.

### Steps to reproduce

1. Create a book from a knowledge base, get a generated proposal.
2. Click "Confirm proposal & build spine".
3. Wait — after ~35-47s the request errors out client-side (spinner/error, or UI appears stuck).
4. Wait another ~30-60s and refresh `My Books` — the book has actually reached `spine_ready` anyway.

### Expected Behavior

The UI shouldn't time out / error on a request that the backend is still successfully processing. Two options that seem reasonable given the existing code:

1. `confirmProposal` (`web/lib/book-api.ts`) is a plain blocking POST that the Next.js middleware (`web/proxy.ts`) rewrites straight through to the backend with no timeout override — but *something* in the path (browser/OS network stack, most likely an idle-connection timeout with no bytes flowing for 35-47s) resets the connection before the multi-LLM spine-build finishes.
2. There's already a WebSocket path for this exact operation — `book.py`'s `/ws` handler accepts a `{"type": "confirm_proposal", "book_id": ..., "proposal": {...}}` message and streams progress back (`confirm_proposal_result`). The "Confirm proposal & build spine" button in `web/app/(workspace)/book/page.tsx` (`handleConfirmProposal`, ~line 287) calls `bookApi.confirmProposal()` (the blocking REST path) instead of using the already-open book WebSocket. Routing this through the WS path (which the frontend already opens for other book events) would sidestep the long-blocking-request-over-HTTP problem entirely and give real progress feedback instead of a dead spinner.

### Related Module

Book Creation

### Configuration Used

LLM: openrouter / anthropic/claude-sonnet-4-5
EMBEDDING: openai / text-embedding-3-large
Deployment: Docker (ghcr.io/hkuds/deeptutor:latest)

### Logs and screenshots

Frontend container log excerpt:
```
INFO:     172.22.0.1:0 - "WebSocket /api/v1/book/ws" [accepted]
Failed to proxy http://localhost:8001/api/v1/book/books/confirm-proposal Error: socket hang up
    at ignore-listed frames {
  code: 'ECONNRESET'
}
Error: socket hang up
    at ignore-listed frames {
  code: 'ECONNRESET'
}
```
No corresponding backend-side error/traceback for the same request in the same log window.

suainam:
## Summary

- route the Web UI confirm-proposal operation through the existing Book WebSocket
- route page compilation through the same streaming transport
- keep the REST endpoints unchanged for CLI and third-party compatibility
- feed streamed operation events into the existing Book progress timeline

## Problem

The current Web UI waits on long-running REST requests. Next.js resets those connections after roughly 35-47 seconds while the FastAPI backend continues generating, so the UI reports a false failure and users may retry the same Book operation.

The same failure mode also affects compile-page during slow providers and fallback chains.

## Fix

The frontend now opens a short-lived Book WebSocket for each long operation, sends the existing confirm_proposal or compile_page message, streams progress events, and resolves only after the matching result event. Each page compilation gets its own socket, so parallel page generation remains parallel.

Browser authentication continues to use the existing dt_token cookie; credentials are not added to the WebSocket URL.

## Verification

- Node regression tests: 161 passed
- ESLint: 0 errors; existing repository warnings unchanged
- Next.js production build: passed
- Exact v1.5.2 backend with only the patched frontend replaced
- confirm_proposal: result received after 41.5 seconds with 9 streamed events
- compile_page: result received after 691.2 seconds with 24 streamed events
- test log audit: 0 REST proxy calls for the affected routes, 0 socket hang ups, 0 ECONNRESET, 0 tracebacks
- production and isolated test containers remained healthy with 0 unexpected restarts

Fixes #607

### DIFF
--- a/web/app/(workspace)/book/page.tsx
+++ b/web/app/(workspace)/book/page.tsx
@@ -14,7 +14,7 @@ import { Loader2, MessageSquare } from "lucide-react";
 import { notify } from "@/lib/notifications";
 import { useTranslation } from "react-i18next";
 
-import { bookApi, openBookSocket } from "@/lib/book-api";
+import { bookApi, type BookWsEvent } from "@/lib/book-api";
 import type {
   Block,
   BlockType,
@@ -132,12 +132,12 @@ function BookPageInner() {
     return () => clearTimeout(timer);
   }, [toast]);
 
-  // ── Live WS event subscription ─────────────────────────────────────
+  // ── Live WS event handling ─────────────────────────────────────────
 
-  useEffect(() => {
-    if (!selectedBookId) return;
-    const socket = openBookSocket((event) => {
-      // Always feed the progress reducer so the timeline updates live.
+  const handleBookOperationEvent = useCallback(
+    (event: BookWsEvent) => {
+      // Each long-running operation owns its WebSocket. Feed those streamed
+      // events into the shared timeline and refresh persisted milestones.
       dispatchProgress(event);
 
       const meta =
@@ -146,23 +146,18 @@ function BookPageInner() {
         (event.content as string) || (meta.kind as string) || "",
       );
       if (
-        kind === "block_ready" ||
-        kind === "block_error" ||
-        kind === "page_compiled" ||
-        kind === "page_planned" ||
-        kind === "spine_ready"
+        selectedBookId &&
+        (kind === "block_ready" ||
+          kind === "block_error" ||
+          kind === "page_compiled" ||
+          kind === "page_planned" ||
+          kind === "spine_ready")
       ) {
         void loadBookDetail(selectedBookId);
       }
-    });
-    return () => {
-      try {
-        socket.close();
-      } catch {
-        // ignore
-      }
-    };
-  }, [selectedBookId, loadBookDetail]);
+    },
+    [selectedBookId, loadBookDetail],
+  );
 
   // ── Selectors ──────────────────────────────────────────────────────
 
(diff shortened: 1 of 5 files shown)


suainam:
Human browser smoke test passed on the isolated v1.5.2 test instance. The UI showed no error, streamed visible progress updates, automatically advanced to Confirm Spine, started compilation after confirmation, and the generated chapter opened successfully. This complements the automated Next.js WebSocket acceptance: confirm_proposal completed in 41.5s and compile_page completed in 691.2s with zero socket hang ups or ECONNRESET.

suainam:
Additional multi-user verification passed after the browser smoke test. Using an enabled non-admin user in the isolated data copy, confirm_proposal completed through the Next.js WebSocket in 43.0s with 10 events, and compile_page completed in 789.0s with 22 events (21 progress events). Both operations exceeded the old HTTP proxy failure window and received their final result frames. The isolated container/data copy was then removed; production remained healthy and unchanged.

pancacake:
Thanks for your contribution!

PR #663, opened 2026-07-22
fix(cli): harden UTF-8 terminal input handling

cmddirexit:
## Summary

- centralize UTF-8-aware reads in the shared CLI console input helper
- temporarily enable `termios.IUTF8` while reading UTF-8 input from a TTY
- restore the original terminal attributes after input completes or is interrupted
- keep the chat REPL alive when malformed terminal bytes still raise `UnicodeDecodeError`
- protect both chat messages and `ask_user` prompts while leaving non-TTY, non-UTF-8, and unsupported platforms unchanged

## Root cause

In canonical mode with `IUTF8` disabled, erasing a multibyte character may remove only its final byte. For example, typing the three-byte UTF-8 character `能` and pressing Backspace can leave the incomplete byte sequence `e8 83`. A strict UTF-8 stdin decoder then raises `UnicodeDecodeError`, terminating the chat REPL.

## Reproduction

1. Run `stty -iutf8`.
2. Start `deeptutor chat` with strict UTF-8 stdin decoding.
3. Type a Chinese character, press Backspace, and submit the line.

## Tests

- added a POSIX pseudo-terminal regression test that starts with `IUTF8` disabled, enters `能`, presses Backspace, and verifies both successful input and terminal-mode restoration
- added a REPL test that injects `UnicodeDecodeError` and verifies the session remains usable
- `python -m pytest -q tests/cli` (41 passed)
- `python -m ruff check deeptutor_cli/chat.py deeptutor_cli/common.py tests/cli/test_chat_cli.py tests/cli/test_chat_terminal.py`
- `python -m ruff format --check deeptutor_cli/chat.py deeptutor_cli/common.py tests/cli/test_chat_cli.py tests/cli/test_chat_terminal.py`

### DIFF
--- a/deeptutor_cli/chat.py
+++ b/deeptutor_cli/chat.py
@@ -2,12 +2,11 @@
 
 from __future__ import annotations
 
-from contextlib import contextmanager, suppress
+from contextlib import suppress
 from dataclasses import dataclass, field
 import json
 import shlex
-import sys
-from typing import Any, Iterator, TextIO
+from typing import Any
 
 from rich.panel import Panel
 import typer
@@ -19,6 +18,7 @@ from .common import (
     maybe_run,
     parse_config_items,
     parse_json_object,
+    read_console_input,
     regenerate_and_render,
     render_tool_result_entry,
     run_turn_and_render,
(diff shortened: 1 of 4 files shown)


pancacake:
Thanks for your contribution!

PR #667, opened 2026-07-23
fix(question): parse quiz JSON with raw_decode

santhreal:
Quiz payloads with trailing prose after the JSON object were dropped to `{}` by greedy brace matching, losing the generated questions.

Parse with `raw_decode` so the leading object is kept.


pancacake:
Thanks for your contribution!

PR #668, opened 2026-07-23
fix(utils): decode trailing-prose JSON before repair

santhreal:
LLM JSON with trailing prose after a valid object failed the primary decode and relied on a weaker repair path.

Try `raw_decode` before repair so the leading object is recovered.


pancacake:
Thanks for your contribution!

PR #669, opened 2026-07-23
fix(visualize): allow code fences without a blank closing line

santhreal:
Fenced code blocks that closed without a blank line before the closing fence were missed, so visualize dropped the code.

Accept closers that do not require a blank preceding line.


pancacake:
Thanks for your contribution!

PR #670, opened 2026-07-23
fix(partners): strip control chars in safe_filename

santhreal:
Partner filenames that contained ASCII control characters survived `safe_filename` and could break downstream path handling.

Strip control characters while keeping the sanitized basename.


pancacake:
Thanks for your contribution!

PR #654, opened 2026-07-20
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- add Atlas Cloud to the shared LLM provider registry as an OpenAI-compatible gateway
- support ATLASCLOUD_API_KEY, the default https://api.atlascloud.ai/v1 endpoint, and atlas/atlas-cloud aliases
- cover provider lookup, runtime config resolution, settings provider choices, and native tool capability behavior

## Validation
- /Users/zby/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -m py_compile deeptutor/services/provider_registry.py deeptutor/services/config/provider_runtime.py tests/services/test_provider_registry.py tests/services/config/test_provider_runtime.py tests/api/test_settings_router.py tests/core/test_agentic_client_provider_kwargs.py
- uv run --python 3.12 --extra dev pytest tests/services/test_provider_registry.py tests/services/config/test_provider_runtime.py tests/api/test_settings_router.py::test_llm_provider_choices_include_atlascloud tests/core/test_agentic_client_provider_kwargs.py::test_registered_cloud_openai_compat_providers_enable_native_tools -q
- git diff --check
- live Atlas Cloud model catalog check returned 425 models and confirmed qwen/qwen3.5-flash and deepseek-ai/deepseek-v4-pro

No README changes; no sponsor, logo, credits, partner, or promotional section changes.

PR #671, opened 2026-07-23
feat: add Eden AI as an LLM provider

MVS-source:
## What

Adds [Eden AI](https://www.edenai.co/) as an LLM provider.

Eden AI is an OpenAI-compatible gateway (`https://api.edenai.run/v3`) that routes to many upstream providers behind a single key. Models are addressed as `<provider>/<model>`, e.g. `openai/gpt-5.5` or `mistral/mistral-large-latest`.

## Changes

A single `ProviderSpec` in `deeptutor/services/provider_registry.py` (per the module's "adding a new provider" note — env vars, config matching and status display all derive from the registry):

- `backend="openai_compat"`, `is_gateway=True`, `env_key="EDENAI_API_KEY"`, `default_api_base="https://api.edenai.run/v3"`.

Placed among the gateways (after OpenRouter), matching the existing order.

## Testing

- `ruff format --check` and `ruff check` pass on the changed file.
- Import smoke test: the registry loads and `edenai` resolves (`canonical_provider_name("edenai") -> "edenai"`).
- Verified live against `https://api.edenai.run/v3`: chat completions return 200 and tool/function calling works for `openai/gpt-5.5` and `mistral/mistral-large-latest`; `GET /models` returns an OpenAI-style list.

Embedding support (Eden AI also exposes `/v3/embeddings`) can follow as a separate change to `EMBEDDING_PROVIDERS` if useful.

Targeting `dev` per CONTRIBUTING.

### DIFF
--- a/deeptutor/services/provider_registry.py
+++ b/deeptutor/services/provider_registry.py
@@ -149,6 +149,16 @@ PROVIDERS: tuple[ProviderSpec, ...] = (
         supports_prompt_caching=True,
     ),
     ProviderSpec(
+        name="edenai",
+        keywords=("edenai", "eden_ai", "eden-ai"),
+        env_key="EDENAI_API_KEY",
+        display_name="Eden AI",
+        backend="openai_compat",
+        is_gateway=True,
+        detect_by_base_keyword="edenai",
+        default_api_base="https://api.edenai.run/v3",
+    ),
+    ProviderSpec(
         name="aihubmix",
         keywords=("aihubmix",),
         env_key="OPENAI_API_KEY",


pancacake:
Merged — thanks for the clean, registry-first addition and the live verification notes. Followed up in 75900a5b: registered the `eden_ai` binding alias in `PROVIDER_ALIASES` (the keywords tuple isn't consulted for gateway name resolution) and backfilled the registry-wiring tests per the Atlas Cloud precedent (#654). Embedding support via `EMBEDDING_PROVIDERS` is welcome as a follow-up PR.

Issue #698, opened 2026-07-25
[Bug]:Assistant replies intermittently disappear from chat; next turn loses context (orphaned branch via stale parent_message_id)


Vincent-LJP:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

In the unified chat workspace, after an assistant reply finishes streaming, the reply sometimes disappears from the conversation view. If I then send another message, the model answers as if it only saw the first turn of the conversation. Switching to another conversation and back makes the missing replies reappear, and subsequent context works again.

## Environment

- DeepTutor version: 1.5.4 (pip install, `deeptutor start`)
- OS: Windows 11
- Browser: Chrome (latest)
- Deployment: local, single user

### Steps to reproduce

<html>
<body>
[...]

PR #706, opened 2026-07-26
fix: persist DONE event ids before a completed turn's execution is popped

### LINKED ISSUE #705: Reconnect after a turn completes drops the DONE event's message ids (root cause of #698)

### Describe the bug

Root cause for #698 (assistant replies intermittently disappear; the next turn loses context).

When a turn finishes normally, `TurnRuntimeManager._run_turn` (`deeptutor/services/session/turn_runtime.py`) publishes the terminal `done` event (carrying `user_message_id` / `assistant_message_id` for the frontend's optimistic-id reconcile) through `_publish_live_event`, which only appends it to the in-memory `_TurnExecution.events` buffer and pushes it to whatever WS subscribers are connected *right now*. The one-shot `_flush_buffered_events` call that persists the buffer to the turn-events store runs *before* that `done` event is constructed, so the `done` event itself, and the id-bearing metadata on it, are never persisted. Once the turn finishes, `_run_turn`'s `finally` block unconditionally pops the turn from `_executions`.

If the client's websocket connection drops during the turn (network blip, a backgrounded/throttled tab, the client's own 45s heartbeat timeout) and reconnects after the turn has already completed and been popped, `subscribe_turn`'s `resume_from` path finds no live execution and no persisted `done` event, so it synthesizes a fallback one (`_synthesize_done_event`) that carries only `{"status": "completed", "synthesized": True}` — no ids. The frontend's `RECONCILE_TURN` dispatch never fires for that turn, so the correctly-persisted assistant message stays outside the client's optimistic-id-reconciled state: the reply appears to disappear, and because the tip id used for the next turn's `parent_message_id` never advances past the previous (already-reconciled) message, the following turn's context effectively resets to before that reply. A full session reload re-fetches the true state from the DB and makes the reply reappear, matching the "switching to another conversation and back" workaround in #698.

### Steps to reproduce

Deterministic backend reproduction (no browser/live LLM needed) — drives `TurnRuntimeManager.start_turn`/`_run_turn`/`subscribe_turn` for real, with only the LLM orchestrator boundary stubbed:

1. Start a turn and let `_run_turn` run to completion **without ever subscribing to it** (simulates a client whose connection was already down when the turn finished; `_run_turn` runs as an independent task regardless of subscribers).
2. Confirm the user+assistant messages are correctly persisted with real ids, and that the turn's execution has been popped from `_executions`.
3. Call `subscribe_turn(turn_id, after_seq=0)` fresh, simulating the client's reconnect.
4. The single `done` event yielded has `metadata == {"status": "completed", "synthesized": True}` — no `assistant_message_id`.

### Expected Behavior

A client reconnecting after a turn has already finished should still receive the persisted `user_message_id` / `assistant_message_id` on `done`, so it can run the same optimistic-id reconcile as a client that stayed connected throughout.

### Related Module

API/Backend

### Configuration Used

N/A — pure backend/session-runtime logic bug, reproduced directly against `deeptutor.services.session.turn_runtime`, unrelated to any specific LLM provider or deployment config.

### Additional Context

Fix + a regression test driving the real `TurnRuntimeManager` (not a mock of the reconcile logic) incoming in a linked PR.

AmirF194:
### Description

Root cause and fix for #698 (assistant replies intermittently disappear; the next turn loses context via a stale parent id). Companion issue: #705.

`_run_turn`'s normal-completion path persisted the buffered turn events to the store *before* constructing and publishing the terminal `done` event. `_publish_live_event` only appends to the in-memory per-turn event buffer and pushes to whatever websocket subscribers are connected at that instant; it never persists anything itself. So the `done` event, carrying `user_message_id`/`assistant_message_id`, was only ever delivered live, never persisted.

Once the turn finishes, its execution is unconditionally popped from `_executions`. A client that reconnects after that (heartbeat timeout, a backgrounded tab, a network blip) calls `subscribe_turn` via the `resume_from` path, finds no live execution and no persisted `done` in the backlog, and gets a synthesized fallback `done` (`_synthesize_done_event`) with no ids. The frontend's `RECONCILE_TURN` dispatch never fires for that turn, so the assistant reply, correctly persisted in the database the whole time, stays outside the client's optimistic-id-reconciled local tree, and the following turn's `parent_message_id` never advances past it. A full session reload re-fetches from the database and the reply reappears, which matches the "switching to another conversation and back" workaround reported in #698.

Fix: move the flush to run once, after the `done` event (and the post-turn `session_meta` title update) have been published, matching the ordering already used on the cancelled and exception paths. `events_flushed` is a single-writer, lock-guarded flag with exactly one `_flush_buffered_events` call site per branch, so this changes what a branch's one flush captures, not how many times it runs.

### Verification

Added `test_reconnect_after_turn_completion_still_carries_message_ids` (`tests/services/session/test_turn_runtime_subscribe.py`), driving the real `TurnRuntimeManager.start_turn` / `_run_turn` / `subscribe_turn` end to end (only the LLM orchestrator boundary is stubbed, same pattern as the existing `test_regenerate.py::test_end_to_end_skips_memory_refresh_and_no_duplicate_user`):

1. Start a turn and let it run to completion with **no subscriber ever attached** (`_run_turn` is an independent task, so it runs regardless).
2. Confirm both messages persisted with real ids and the execution was popped from `_executions`.
3. Call `subscribe_turn(turn_id, after_seq=0)` fresh (the reconnect path) and assert the yielded `done` event's metadata carries the real `assistant_message_id`.

Confirmed this fails on `dev` HEAD (`{"status": "completed", "synthesized": True}`, no id) and passes on this branch, both in a clean `python:3.13-slim` Docker container with an editable install. Also ran the full `tests/services/session/` suite (153 tests) and `tests/api/test_unified_ws_turn_runtime.py` in the same container: all pass, no regressions.

Updated a stale docstring in `test_turn_event_flush.py` that documented the old (buggy) "flush right before DONE" ordering.

`ruff check` and `ruff format --check` clean on the changed files (this repo's CI only runs Ruff for Python; I did not run `mypy`/`bandit`/`interrogate`, which are configured in `pre-commit` but are not invoked by any `.github/workflows/*.yml` job). Did not run the Web Node Tests job since no frontend files changed.

**Invariant check** (`events_flushed` / `_executions`, both shared mutable state): `events_flushed` has exactly one writer (`_flush_buffered_events`, under `self._lock`); `_executions` has exactly one insertion site (`start_turn`) and one removal site (`_run_turn`'s `finally`), both under the same lock. No test in this PR mocks any of those three sites: the real lock, the real dict, and the real flag are exercised.

### Related Issues
- Fixes #705
- Root cause of #698

### Module(s) Affected
- [x] `services`
- [x] `tests`

### Checklist
- [x] I have read and followed the contribution guidelines.
- [x] `ruff check` / `ruff format --check` clean (see Verification).
- [ ] I have updated the documentation (if necessary): none needed, internal runtime behavior only.

### Additional Notes

Targeting `dev` per CONTRIBUTING.md (general bug fix, not multi-user-specific).

### DIFF
--- a/deeptutor/services/session/turn_runtime.py
+++ b/deeptutor/services/session/turn_runtime.py
@@ -1766,7 +1766,6 @@ class TurnRuntimeManager:
                     events=assistant_events,
                     attachments=generated_attachments or None,
                 )
-            await self._flush_buffered_events(execution)
             turn_status, turn_error = _resolve_turn_outcome(
                 assistant_events,
                 pending_done_event,
@@ -1812,6 +1811,17 @@ class TurnRuntimeManager:
                     )
                 except Exception:
                     logger.debug("Failed to generate session title", exc_info=True)
+            # Flush once every terminal/post-turn event (DONE, and the title
+            # ``session_meta`` above) has been published, not before: a
+            # client that reconnects after this task's ``finally`` pops
+            # ``execution`` from ``_executions`` falls back entirely to this
+            # persisted backlog, and ``subscribe_turn`` synthesises an
+            # id-less DONE when it finds none there -- permanently orphaning
+            # the just-persisted assistant reply from that client's
+            # reconcile path (it can still see the message after a full
+            # session reload, since the row itself is fine; only the
+            # targeted in-place swap is unreachable).
+            await self._flush_buffered_events(execution)
         except asyncio.CancelledError:
             if not stream_done_sent:
                 await self._publish_live_event(
--- a/tests/services/session/test_turn_event_flush.py
+++ b/tests/services/session/test_turn_event_flush.py
@@ -1,10 +1,11 @@
 """Post-stream turn-event flush: batching, workspace mirror, PocketBase upload.
 
 The turn runtime buffers every live event in memory and persists the whole
-batch after the stream drains, right before publishing DONE. Everything on
-that path must stay O(1) round-trips w.r.t. the event count — per-event
-commits/opens/POSTs sat between the last streamed token and the client's
-spinner clearing (the "stuck on generating" report).
+batch, including DONE and any post-turn ``session_meta`` title update, once
+the turn has fully finished. Everything on that path must stay O(1) round
+trips w.r.t. the event count — per-event commits/opens/POSTs sat between the
+last streamed token and the client's spinner clearing (the "stuck on
+generating" report).
 """
 
 from __future__ import annotations
--- a/tests/services/session/test_turn_runtime_subscribe.py
+++ b/tests/services/session/test_turn_runtime_subscribe.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+from types import SimpleNamespace
 
 import pytest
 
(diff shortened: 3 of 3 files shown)


AmirF194:
No rush, just a check-in on this one. CI is green and it's a small, self-contained fix if that helps with prioritizing.

PR #715, opened 2026-07-27
fix: parse_language() collapses non-English codes to Chinese

### LINKED ISSUE #712: [Bug]: parse_language() collapses every non-English language code to Chinese (ja/ko/de/fr/... -> zh)

### Describe the bug

`parse_language()` in `deeptutor/services/config/loader.py` is a **binary zh/en classifier**: any language code other than `en` falls through to a hardcoded `return "zh"`. Requesting Japanese (`ja`) therefore silently produces **Chinese** output.

I hit this while using DeepTutor v1.5.4 in Japanese. Book chapter prose was correctly generated in Japanese, but the **quizzes came back in Chinese**, mixed into an otherwise Japanese book.

The cause is the final line of the parser:

https://github.com/HKUDS/DeepTutor/blob/main/deeptutor/services/config/loader.py#L171-L195

```python
def parse_language(language: Any) -> str:
    if not language:
        return "zh"
    if isinstance(language, str):
        lang_lower = language.lower()
        if lang_lower in ["en", "english"]:
            return "en"
        if lang_lower in ["zh", "chinese", "cn"]:
            return "zh"
    return "zh"  # Default Chinese
```

Verified by executing the function directly:

```
parse_language('ja')       -> 'zh'
parse_language('japanese') -> 'zh'
parse_language('ko')       -> 'zh'
parse_language('fr')       -> 'zh'
parse_language('de')       -> 'zh'
parse_language('es')       -> 'zh'
parse_language('ru')       -> 'zh'
parse_language('ar')       -> 'zh'
parse_language('en')       -> 'en'
parse_language('zh')       -> 'zh'
```

**Every language except English resolves to Chinese.** This is not specific to Japanese.

#### Why only some features are affected

The codebase has **two parallel language resolution paths**, and only one of them is correct:

| Path | Behavior |
|:---|:---|
| `deeptutor/services/prompt/language.py` | ✅ Correct — `_LANGUAGE_LABELS` already maps `ja`, `ko`, `es`, `fr`, `de`, `ru`, `pt`, `it` |
| `deeptutor/services/config/loader.py` `parse_language()` | ❌ Binary zh/en only |

Book prose goes through the first path, so it renders in Japanese. Quiz generation goes through the second, so it renders in Chinese. That is exactly the mixed-language output I observed.

#### Affected call sites

`parse_language()` has four downstream consumers, so the impact is wider than quizzes alone:

| Call site | Affected feature |
|:---|:---|
| `deeptutor/agents/question/pipeline.py:370` | **Quiz / Question generation** (confirmed in practice) |
| `deeptutor/agents/research/pipeline.py:284` | **Deep Research** |
| `deeptutor/learning/prompts.py:33` | **Mastery Path / Guided Learning** |
| `deeptutor/services/prompt/manager.py:73,216` | **Prompt loading generally** |

`deeptutor/learning/prompts.py` compounds the problem with a second Chinese-biased fallback:

```python
lang = parse_language(language)
candidates = [lang, "zh" if lang != "zh" else "en"]
```

so an unsupported language falls back to Chinese there too, rather than to English.

#### Relation to #33

Issue #33 (*[architectural debt] refactor fragmented prompt loading and hardcoded language logic*) reported this same class of problem — explicitly calling out *"forcing a fallback to `zh`"* — and was closed as addressed in v0.3.0. The `parse_language()` fallback appears to have survived that refactor, so this looks like a **regression / incomplete fix** of #33 rather than a new architectural concern.

### Steps to reproduce

1. Configure any LLM provider (I used DeepSeek `deepseek-v4-pro`).
2. Generate a Book with Japanese content — e.g. via the API with `language: "ja"`, or simply prompt in Japanese.
3. Open a chapter containing a `quiz` block.

**Result:** chapter prose is in Japanese, but the quiz questions, options and explanations are in Chinese.

Minimal reproduction without running the server:

```python
from deeptutor.services.config import parse_language
assert parse_language("ja") == "ja"   # fails — returns "zh"
```

### Expected Behavior

`parse_language("ja")` should resolve to Japanese, and reader-facing text in Quiz / Deep Research / Mastery Path should be generated in the requested language.

More generally:

1. `parse_language()` should not collapse arbitrary language codes to `zh`. It could delegate to the existing, already-correct `normalize_language()` / `language_label()` in `deeptutor/services/prompt/language.py` instead of maintaining a second, weaker implementation.
2. When a language genuinely has no localized prompt YAML, the fallback should be **English**, not Chinese — matching the `"en"` default used elsewhere (e.g. `UISettings`).

Note that this does **not** require adding new translated prompt resources: the LLM produces the text, so simply passing the correct language code through is enough for the model-facing paths. This makes it a much smaller change than a full locale addition.

### Related Module

API/Backend

### Configuration Used

```
LLM: DeepSeek deepseek-v4-pro
Embedding: not configured
KB: not used
```

### Logs and screenshots

_No response_

### Additional Information

- AI-Tutor Version: v1.5.4 (source install)
- Operating System: Windows 11
- Python Version: 3.11
- Node.js Version: 20
- Browser (if applicable): —
- Related Issues: #33 (closed, same class of problem), #684, #471

I'm happy to open a PR for this if the maintainers are open to it — the change is confined to `parse_language()` and the `learning/prompts.py` fallback, and adds no new locale files.

kagura-agent:
## Problem

`parse_language()` in `loader.py` is a binary zh/en classifier — any language code other than `"en"` (including `"ja"`, `"ko"`, `"fr"`, `"de"`, etc.) silently returns `"zh"`. This causes mixed-language output: for example, Japanese book prose renders correctly, but quizzes come back in Chinese.

```python
parse_language('ja')  # → 'zh' (wrong!)
parse_language('ko')  # → 'zh' (wrong!)
parse_language('fr')  # → 'zh' (wrong!)
```

## Root Cause

`parse_language()` only recognizes `"en"` and `"zh"`, with a hardcoded `return "zh"` fallback. A correct multi-language implementation exists in `language.py` (`_LANGUAGE_LABELS`), but `parse_language()` doesn't use it.

### 1. `deeptutor/services/config/loader.py`
- Rewrite `parse_language()` with an alias table for common full names (`"japanese"→"ja"`, `"korean"→"ko"`, etc.) matching the codes already in `_LANGUAGE_LABELS`
- Pass through unknown ISO 639 codes lowercased (forward-compatible)
- Change default from `"zh"` to `"en"` to align with `config.system.language` and `UISettings` defaults

### 2. `deeptutor/services/prompt/manager.py`
- Improve fallback chain for unknown codes: try the requested language's prompt directory before falling back to English (`[lang_code, "en"]` instead of `["en"]`)

### 3. `tests/services/config/test_parse_language.py` (new)
- 31 parametrized tests covering aliases, ISO codes, edge cases (None, empty, non-string, whitespace), and regional codes

## Design Notes

- **No import from `language.py`** in `loader.py` to avoid circular imports (`config` is a lower-level module imported by `prompt`)
- **Default change zh→en**: Aligns with the system's own defaults. Users who explicitly set `"zh"` are unaffected
- **Pass-through for unknown codes**: Acceptable because `PromptManager` already falls back to `"en"` when no prompt YAML exists for a code
- Downstream `== "zh"` checks: non-Chinese users correctly get English templates, while `append_language_directive()` handles LLM output language

Closes #712

### DIFF
--- a/deeptutor/services/config/loader.py
+++ b/deeptutor/services/config/loader.py
@@ -168,31 +168,36 @@ def get_path_from_config(config: dict[str, Any], path_key: str, default: str = N
     return default
 
 
+_LANGUAGE_ALIASES = {
+    "english": "en",
+    "chinese": "zh",
+    "cn": "zh",
+    "japanese": "ja",
+    "korean": "ko",
+    "spanish": "es",
+    "french": "fr",
+    "german": "de",
+    "russian": "ru",
+    "portuguese": "pt",
+    "italian": "it",
+}
+
+
 def parse_language(language: Any) -> str:
     """
     Unified language configuration parser, supports multiple input formats
 
-    Supported language representations:
-    - English: "en", "english", "English"
-    - Chinese: "zh", "chinese", "Chinese"
-
     Args:
-        language: Language configuration value (can be "zh"/"en"/"Chinese"/"English" etc.)
+        language: Language configuration value (ISO 639 code or full name)
 
     Returns:
-        Standardized language code: 'zh' or 'en', defaults to 'zh'
+        Standardized language code, defaults to 'en'
     """
-    if not language:
-        return "zh"
+    if not isinstance(language, str) or not language.strip():
+        return "en"
 
-    if isinstance(language, str):
-        lang_lower = language.lower()
-        if lang_lower in ["en", "english"]:
-            return "en"
-        if lang_lower in ["zh", "chinese", "cn"]:
-            return "zh"
-
-    return "zh"  # Default Chinese
+    lang_lower = language.strip().lower()
+    return _LANGUAGE_ALIASES.get(lang_lower, lang_lower)
 
 
 def get_agent_params(module_name: str) -> dict:
--- a/deeptutor/services/prompt/manager.py
+++ b/deeptutor/services/prompt/manager.py
@@ -100,7 +100,7 @@ class PromptManager:
     ) -> dict[str, Any]:
         """Load prompt file with language fallback."""
         prompt_dirs = self._candidate_prompt_dirs(module_name)
-        fallback_chain = self.LANGUAGE_FALLBACKS.get(lang_code, ["en"])
+        fallback_chain = self.LANGUAGE_FALLBACKS.get(lang_code, [lang_code, "en"])
 
         for prompts_dir in prompt_dirs:
             for lang in fallback_chain:
--- a/tests/services/config/test_parse_language.py
+++ b/tests/services/config/test_parse_language.py
@@ -1,0 +1,43 @@
+import pytest
+
+from deeptutor.services.config.loader import parse_language
+
+
+@pytest.mark.parametrize(
+    "input_val, expected",
+    [
+        ("en", "en"),
+        ("English", "en"),
+        ("ENGLISH", "en"),
+        ("zh", "zh"),
+        ("Chinese", "zh"),
+        ("cn", "zh"),
+        ("CN", "zh"),
+        ("ja", "ja"),
+        ("ko", "ko"),
+        ("fr", "fr"),
+        ("de", "de"),
+        ("es", "es"),
+        ("ru", "ru"),
+        ("pt", "pt"),
+        ("it", "it"),
+        ("Japanese", "ja"),
+        ("Korean", "ko"),
+        ("French", "fr"),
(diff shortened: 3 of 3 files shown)


kagura-agent:
Closing — this was fixed directly on `dev` in 59132932 by @pancacake. Glad the analysis was helpful! 🙏

PR #716, opened 2026-07-27
fix: guide remote Codex OAuth loopback callbacks

TyrionH-is-coding:
## Summary

- keep the Codex OAuth listener loopback-only while forwarding only the exact browser callback path through the DeepTutor Web origin to a state-validating broker
- detect remote Web access and show the real callback address, SSH tunnel command, explicit authorization action, countdown, and refresh-safe operation state
- align CLI output and English/Chinese deployment docs, including Docker host-port and reverse-proxy caveats

## Root cause

For a remote deployment, the browser's `localhost` is the user's workstation while the OAuth listener runs on the server (and, under Docker, inside the container). A normal reverse proxy cannot deliver `http://localhost:1455|1457/auth/callback` to that private listener.

## Security and behavior

- the listener remains loopback-only; ports `1455` and `1457` are not published
- the Web bridge rewrites only `/auth/callback`
- the broker requires an active operation and constant-time matching OAuth `state`
- callback responses use `no-store` HTML and never echo the authorization code or state
- PKCE and the existing token exchange remain unchanged

## Validation

- targeted Python OAuth/API/CLI tests: 62 passed locally on the current head; the same implementation passed the equivalent temporary Linux container suite before this cleanup-only commit
- `npm run test:node`: 258 passed
- `npm run i18n:check`
- `npx tsc --noEmit`
- `npm run build`
- `npm run lint`: 0 errors (existing warnings remain)
- Ruff 0.16 check and format check
- temporary remote Docker acceptance: wrong state was rejected without terminating the operation; a valid-state callback traversed SSH tunnel → Web rewrite → broker → loopback listener and returned redacted HTTP 200; active guidance recovered after a page reload with a decreasing countdown

The remote topology test used a dummy authorization code, so it intentionally stopped at the unchanged token-exchange step rather than consuming a real account authorization.


pancacake:
Merged into `dev` as 1117e6fe (squashed by hand — this branch conflicted with #713, which rewrote the `codex_auth` service's owner resolution after you opened this, so `gh pr merge` couldn't do it. The conflict was only the import block: `get_path_service` is gone from module scope on `dev`, so the merged imports keep just `load_system_settings`).

Thanks for this — the listener staying loopback-only, the single rewritten path, the constant-time state match, the `no-store` responses that never echo the code, and binding with `start_serving=False` to close the accept-before-init window all made this easy to review.

One correctness gap I fixed on top, in 7a5102a8, because it made the feature fail for exactly the deployments it targets:

`GET /api/v1/auth/openai-codex/callback` cannot be authenticated — the browser reaches it on the far end of the user's own tunnel, a different origin from the DeepTutor Web session, so no `dt_token` cookie is sent. But the current-user ContextVar is only ever set inside `require_auth` (`deeptutor/api/routers/auth.py:230`), so on that route it stays unset and `get_current_path_service()` falls back to `PathService.get_instance()`. `get_codex_oauth_service()` keys its instance registry on the resolved user root, and a non-administrator's root is `data/users/<id>` while the default is `data/user` — so the login operation and the callback landed on two different service instances, and every non-admin remote sign-in got `login_not_active` (409). The endpoint tests didn't catch it because they monkeypatch the factory with a fake, so user resolution was never exercised.

The fix routes delivery by OAuth `state` instead of by request user: the state is a secret this process minted for exactly one login and is already compared in constant time, which is the same trust model your loopback listener uses, so no new authorization surface opens. The endpoint's status codes are unchanged (409 when nothing is waiting, 400 when a login is waiting but the state doesn't match), the endpoint tests now patch the delivery seam, and three service-level tests cover routing to a non-default instance.

Gates on the merged result: `ruff check` + `ruff format --check` clean, 3010 passed / 8 skipped, `tsc --noEmit` clean, `eslint .` 0 errors, `i18n:check` clean, `test:node` 258 passed.

PR #726, opened 2026-07-30
feat(settings): expose language switcher

junlytech-ship-it:
## What

- Add a discoverable English / Simplified Chinese selector to the main Settings header.
- Keep the control responsive and accessible with a localized group label and `aria-pressed` state.
- Reuse the existing `SettingsContext` language state and persistence action.

## Why

DeepTutor already supports English and Simplified Chinese, but the selector is currently nested under Appearance. Exposing it on the Settings landing page makes the bilingual experience much easier to discover, especially for Chinese-speaking users.

## Scope

This is a frontend-only settings enhancement. It does not change the existing Agent, Capability, partner, or task orchestration architecture.

## Validation

- 272 Node tests passed
- Targeted ESLint passed
- i18n parity check passed
- Next.js production build passed
- Production Docker image built successfully
- Container health check passed and `/settings` returned HTTP 200
- Browser verified English/Chinese switching and persisted Chinese selection
- Browser verified the Settings layout at desktop and 768 px viewport widths

### DIFF
--- a/web/components/settings/LanguageSwitcher.tsx
+++ b/web/components/settings/LanguageSwitcher.tsx
@@ -1,0 +1,54 @@
+"use client";
+
+import { Languages } from "lucide-react";
+
+import type { Lang } from "@/lib/settings-nav";
+
+type LanguageSwitcherProps = {
+  language: "en" | "zh";
+  onChange: (language: "en" | "zh") => void;
+  tr: (label: Lang) => string;
+};
+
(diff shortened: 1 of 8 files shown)


pancacake:
Thanks for your comments! Closing this one as superseded: #758 landed the same Settings language control as part of a wider change that also splits the interface locale from the model-output language (#757), so the switcher here would now conflict with it rather than add to it.

PR #765, opened 2026-08-04
feat: mark Kimi K3 as vision capable

chiruu12:
## Summary

`kimi-k2.5` and `kimi-k2.6` are listed as vision capable but `kimi-k3` is not, so it reported `supports_vision=False` and image input was skipped for it. K3 is natively multimodal.

One entry alongside the existing Kimi ones, plus a test. Ruff and format are clean.


pancacake:
Merged into `dev` as 6c7ee227 — rebased locally to resolve the append conflict in `tests/services/llm/test_capabilities.py` against #764, which landed first. Your commit and authorship are preserved verbatim. Thanks!

PR #785, opened 2026-08-06
[Bug]:Frontend proxy returns 500 on Windows local deployment due to localhost IPv6 resolution

### LINKED ISSUE #782: [Bug]:Frontend proxy returns 500 on Windows local deployment due to localhost IPv6 resolution

### Describe the bug

When deploying locally on Windows (non‑Docker) using `deeptutor start`, the frontend loads but all proxied `/api/*` requests return HTTP 500. The exact error displayed in the frontend is:
“Could not load settings from the backend. Settings fetch failed: HTTP 500. Verify the backend is running and NEXT_PUBLIC_API_BASE points to a reachable host.”
Direct access to the backend at `http://127.0.0.1:8001` works (returns 200), so this is a proxy‑related failure.
I searched existing issues and didn’t find this reported. I believe this is a genuine bug in the launcher code, not a configuration or user‑error issue, because it stems from hardcoded logic that overrides user settings and fails specifically on Windows due to IPv6 resolution.

### Steps to reproduce

1. On a Windows machine (non‑Docker), clone the repository and set up the project.
2. Run `deeptutor start` to start both frontend and backend.
3. Open the frontend in a browser (e.g., Chrome).
4. Observe that the frontend page loads, but any API call (e.g., loading settings) fails with a 500 error.
5. Manually access `http://127.0.0.1:8001` (or `http://localhost:8001` directly) – it returns a successful 200 response.

### Expected Behavior

The frontend proxy should forward requests to the backend successfully, and the settings should load without errors.

### Related Module

API/Backend

### Configuration Used

  OS: Windows 11 家庭版 中文版 (25H2, build 26200.8655)
  Processor: Intel Core Ultra 5 125H
  RAM: 32.0 GB
  DeepTutor version: latest
  Python version: 3.13
  Node.js version: v18.20.8
  Browser: Microsoft Edge
  Deployment type: Local, non-Docker, started via deeptutor start

### Logs and screenshots

I have verified that changing `localhost` to `127.0.0.1` in `launcher.py` immediately resolves the issue, confirming that the fix targets the root cause.
Additionally, to confirm reproducibility, I reproduced the 500 error multiple times before applying the fix (even after restarting the service), so it is not an intermittent issue.

### Additional Information

Root Cause:
1. runtime/launcher.py hardcodes backend_url = f"http://localhost:{backend_port}" and overrides DEEPTUTOR_API_BASE_URL, ignoring the next_public_api_base setting from system.json.
2. Node.js (Next.js middleware / undici) resolves localhost to IPv6 ::1 first, so fetch('http://localhost:8001') fails with ECONNREFUSED.
3. The uvicorn backend is started with --host 0.0.0.0 (IPv4 only), so the IPv6 connection is refused, leading to 500.

Suggested Fixes:
- Use http://127.0.0.1:{port} in the launcher, or
- Start uvicorn with --host :: (dual-stack), or
- Respect the configured next_public_api_base when generating DEEPTUTOR_API_BASE_URL.

Temporary Workaround (currently used):
I have resolved the issue locally by patching the two occurrences of localhost to 127.0.0.1 in launcher.py. This is my current workaround.

Impact Scope:
I have not used Docker deployment, so I have not tested it. However, theoretically, Docker deployments should be unaffected because container-internal localhost resolution behaves differently. Therefore, this issue is likely specific to Windows native (non-Docker) deployment.

loonister1:
The launcher now hands the Next.js middleware an IPv4 loopback backend URL (`http://127.0.0.1:{port}`) instead of `http://localhost:{port}`, so Node's `::1`-first resolution no longer collides with uvicorn's IPv4-only `0.0.0.0` bind and `/api/*` rewrites stop failing with ECONNREFUSED → HTTP 500. A `_loopback_url` helper carries the reasoning and gives the fix a testable seam; `web/proxy.ts`'s own no-env fallback was aligned for the same reason, and its stale note about what `deeptutor start` exports was corrected.

Fixes #782

### Testing
[targeted; 3 commit(s) checked individually] $ /Users/m2/Documents/PW/github-bot/work/HKUDS__DeepTutor/782/.contrib-venv/bin/python -m pytest -q --no-header tests/runtime/test_launcher.py

### Notes for reviewers
Scrutinise the test first: `assert launcher._loopback_url(8001) == "http://127.0.0.1:8001"` restates the helper's one-line body and never exercises `start()`, so reverting either call site to `f"http://localhost:{backend_port}"` leaves it green — it gives no regression cover for the defect it names. If `start()` is reachable under monkeypatch, ask for an assertion on the exported `common_env["DEEPTUTOR_API_BASE_URL"]` instead; if it is not (it spawns processes), the current test is the practical ceiling and worth accepting. Second, the diff truncates `api_base = (runtime_env.get("NEXT_PUBLIC_API_BASE_EXTERNAL") or runtime_env.get("NEXT_PUBLIC_API_BASE") or ...`; if the final fallback is `backend_url`, this silently changes the browser-facing `NEXT_PUBLIC_API_BASE` default from `localhost` to `127.0.0.1`, which is a distinct origin for cookies and for any backend CORS allowlist that hardcodes `http://localhost:*` — check those two before merging. Third, the Docker entrypoint still falls back to `http://localhost:${BACKEND_PORT}`, so proxy.ts's comment now documents two different fallbacks; containers with `::1 localhost` in /etc/hosts can hit the same undici ordering, and whether to align the entrypoint is your call, not a defect in this patch. Fourth, confirm the uvicorn bind host is not user-configurable; if it is, hardcoding 127.0.0.1 rather than deriving the loopback from the configured bind is a design decision that belongs to you. Finally, the patch deliberately does not adopt the issue's third suggestion (respecting `next_public_api_base` when generating `DEEPTUTOR_API_BASE_URL`) — the pre-existing comment justifies that separation, but the reporter will expect an explanation.

### DIFF
--- a/deeptutor/runtime/launcher.py
+++ b/deeptutor/runtime/launcher.py
@@ -793,6 +793,17 @@ def _coerce_port(value: object) -> int | None:
     return port if 1 <= port <= 65535 else None
 
 
+def _loopback_url(port: int) -> str:
+    """Return the loopback URL for ``port``, pinned to IPv4.
+
+    Node resolves ``localhost`` to ``::1`` before ``127.0.0.1`` while uvicorn
+    binds ``0.0.0.0`` (IPv4 only), so a ``localhost`` proxy target left every
+    ``/api/*`` rewrite refused -> HTTP 500 in the browser (issue #782).
+    """
+
+    return f"http://127.0.0.1:{port}"
+
+
 def _local_app_url(value: object, port: int) -> str:
     fallback = f"http://localhost:{port}"
     if not isinstance(value, str) or not value.strip():
@@ -955,7 +966,7 @@ def start(home: str | Path | None = None, *, dev: bool = False) -> None:
 
     backend_port = settings.backend_port
     frontend_port = settings.frontend_port
-    backend_url = f"http://localhost:{backend_port}"
+    backend_url = _loopback_url(backend_port)
     api_base = (
         runtime_env.get("NEXT_PUBLIC_API_BASE_EXTERNAL")
         or runtime_env.get("NEXT_PUBLIC_API_BASE")
@@ -991,7 +1002,7 @@ def start(home: str | Path | None = None, *, dev: bool = False) -> None:
     if (resolved_backend, resolved_frontend) != (backend_port, frontend_port):
         backend_port, frontend_port = resolved_backend, resolved_frontend
         runtime_env = export_runtime_settings_to_env(overwrite=True)
-        backend_url = f"http://localhost:{backend_port}"
+        backend_url = _loopback_url(backend_port)
         api_base = (
             runtime_env.get("NEXT_PUBLIC_API_BASE_EXTERNAL")
             or runtime_env.get("NEXT_PUBLIC_API_BASE")
@@ -1032,7 +1043,7 @@ def start(home: str | Path | None = None, *, dev: bool = False) -> None:
     # The Next.js middleware (web/proxy.ts) runs in the frontend's Node runtime
     # and reads these at request time to forward /api/* and /ws/* to the backend
     # and to gate the login redirect. The browser uses relative paths, so the
-    # frontend server reaches the backend on localhost at the resolved port —
+    # frontend server reaches the backend on the loopback at the resolved port —
     # use backend_url (not api_base, which may be an external browser URL).
     common_env["DEEPTUTOR_API_BASE_URL"] = backend_url
     common_env["DEEPTUTOR_AUTH_ENABLED"] = "true" if auth_enabled else "false"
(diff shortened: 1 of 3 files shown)


pancacake:
Thank you for picking this up — your diagnosis matches ours exactly, and the reasoning in your `_loopback_url` docstring is the right one.

This landed on `dev` before the PR arrived, so it is already covered:

- `110dd902` (#784) — `backend_url` pinned to `http://127.0.0.1:{backend_port}` at both sites in `launcher.py`.
- `9af5f890` — the same in `web/proxy.ts`, including the `http://localhost:8001` fallback your diff also fixes.
- `tests/runtime/test_launcher.py::test_start_uses_ipv4_loopback_for_frontend_proxy` covers the regression by driving `start()` and asserting the exported `DEEPTUTOR_API_BASE_URL`, so it holds for the whole resolution path rather than one helper.

Closing as already-fixed rather than as a rejection — the change was correct, we just got there first. Issue #782 is closed too. Please do keep them coming.

PR #800, opened 2026-08-08
fix: resolve generated-output downloads per user

### LINKED ISSUE #790: [Bug]:Multi-user File Download Fails with 404 "Output not found" on /api/outputs
# GitHub Issue: Multi-user File Download Fails with 404 "Output not found" on /api/outputs

## Issue Title
`[Bug] Multi-user file download fails with 404 (Output not found) on /api/outputs static mount`

---

## 🐛 Bug Description
In a multi-user deployment of DeepTutor, when a non-admin user (e.g., `xxx` under UID `xxxxxx`) requests the generation of files (such as `.docx` or `.pdf` documents) in the chat, the backend successfully generates and saves the files in the user's isolated workspace:
`data/users/<uid>/user/workspace/...`

However, attempting to download these files via the web interface by requesting the static route:
`GET /api/outputs/workspace/chat/chat/...`
consistently fails with a `404 Not Found` response returning `{"detail":"Output not found"}`.

---

## 🔍 Bug or Configuration?
This is **100% a code logic bug** within the static files serving architecture, not a configuration issue. 
No configuration in `system.json`, `interface.json`, or environment variables can fix this.

### Root Cause Analysis
1. **The Startup Bind**: In `deeptutor/api/main.py`, the static file mount `/api/outputs` is registered at FastAPI startup using `SafeOutputStaticFiles`:
   ```python
   path_service = get_path_service()
   user_dir = path_service.get_public_outputs_root()
   
   app.mount(
       "/api/outputs",
       SafeOutputStaticFiles(directory=str(user_dir), path_service=path_service),
       name="outputs",
   )
   ```
   At startup, there is no active HTTP request context. Thus, `get_path_service()` falls back to the default `PathService` instance, hardcoding `user_dir` to `/data/user`.

2. **Static Directory Hardcoding**: FastAPI/Starlette's `StaticFiles` constructor binds the `self.directory` to the provided `user_dir` string during startup. When a request later comes in, `StaticFiles` attempts to resolve files relative to this hardcoded path (`data/user/`).

3. **Multi-User Isolation**: In a multi-user context, files are saved under `data/users/<uid>/user/`. When `SafeOutputStaticFiles.get_response` receives a request, it calls `super().get_response(path, scope)`. Starlette then looks for the file inside `data/user/workspace/...` instead of `data/users/<uid>/user/workspace/...`, resulting in a file-not-found error (or path-service validation failure) and returning `404 Output not found`.

---

## 🏃 Steps to Reproduce
1. Enable multi-user authentication in DeepTutor.
2. Sign in as a regular user (e.g. `xxxx`).
3. Send a prompt to generate a file (e.g. *"Generate a python script and output it as a PDF"*).
4. Wait for the pipeline to write the file under `data/users/<uid>/user/workspace/chat/chat/turn_.../code_runs/.../file.pdf`.
5. Click the download link generated in the chat window, or perform a manual `GET` request:
   `curl -i http://<host>:3782/api/outputs/workspace/chat/chat/turn_.../code_runs/.../file.pdf`
6. Observe the response: `HTTP/1.1 404 Not Found` with `{"detail":"Output not found"}`.

---

## 💡 Proposed Architectural Fix
Rewrite `SafeOutputStaticFiles.get_response` inside `deeptutor/api/main.py`. Since static mounts run as independent ASGI sub-apps, standard request middleware (which sets the request-local `ContextVar` for the logged-in user) is bypassed or lost due to context propagation limits. 

Therefore, we must manually parse the `dt_token` from the request headers' Cookie, decode it, resolve the user's scope PathService, and fallback to the default PathService when no session is present.

```python
class SafeOutputStaticFiles(StaticFiles):
    """Static file mount that dynamically resolves and exposes whitelisted user artifacts."""

    def __init__(self, *args, path_service, **kwargs):
        super().__init__(*args, **kwargs)
        self._path_service = path_service  # Kept as fallback/compatibility

    async def get_response(self, path: str, scope):
        from fastapi.responses import FileResponse
        from deeptutor.services.path_service import PathService
        from deeptutor.multi_user.context import user_from_token_payload
        from deeptutor.multi_user.paths import get_path_service_for_scope
        from deeptutor.api.routers.auth import decode_token
        from fastapi import HTTPException

        # 1. Manually parse the "dt_token" credential from the Request Cookie header
        headers = dict(scope.get("headers", []))
        cookie_bytes = headers.get(b"cookie", b"")
        cookie_str = cookie_bytes.decode("utf-8", errors="ignore")
        
        token = None
        for item in cookie_str.split(";"):
            item = item.strip()
            if item.startswith("dt_token="):
                token = item.split("=", 1)[1]
                break

        # 2. Decode the JWT and resolve the context PathService for the user scope
        current_path_service = None
        if token:
            try:
                payload = decode_token(token)
                if payload:
                    user = user_from_token_payload(payload)
                    current_path_service = get_path_service_for_scope(user.scope)
            except Exception:
                pass

        # 3. Fallback: If no token or invalid, default to the global PathService
        if current_path_service is None:
            from deeptutor.services.path_service import get_path_service
            current_path_service = get_path_service()

        # 4. Safety Check: Verify if the requested path matches the whitelist
        if not current_path_service.is_public_output_path(path):
            raise HTTPException(status_code=404, detail="Output not found")

        # 5. Resolve the actual physical absolute path in the user's isolated workspace
        physical_file_path = (current_path_service.get_public_outputs_root() / path).resolve()

        # 6. Return the file safely using FileResponse
        if not physical_file_path.is_file():
            raise HTTPException(status_code=404, detail="Output file not found on disk")

        return FileResponse(str(physical_file_path))
```

---

## 💻 Environment
- **DeepTutor version**: v1.5.8 (with multi-user schema enabled)
- **Framework**: FastAPI / Starlette
- **OS**: Linux (Ubuntu 24.04)

---

## 🚀 Verified Solution Implementation
This dynamic routing fix has been successfully implemented, deployed, and verified in our environment via custom patches. 

Regular users (e.g., `xxx` under UID `xxxxx`) are now able to successfully download generated `.docx` and `.pdf` files without any `404` errors. Meanwhile, security rules (e.g., directory traversal protection, private suffix filtering) remain fully intact and function correctly.

mvanhorn:
### Description

Update `SafeOutputStaticFiles` in `deeptutor/api/main.py` to select a `PathService` for each request instead of delegating lookup to the startup-bound `StaticFiles.directory`: read the `dt_token` cookie from the ASGI request, decode it through the existing auth service, convert the payload through the existing multi-user identity mapping, and resolve that scope with `get_path_service_for_scope`. Resolve and serve the requested file against that service's public-output root only after `is_public_output_path` accepts it, so validation and physical lookup use the same account root. Preserve the default service only for auth-disabled/local-admin operation; when authentication is enabled, a missing or invalid credential should return the same non-revealing 404 rather than falling back to another account's files.

`SafeOutputStaticFiles` is constructed at application startup with the default `PathService`, so its Starlette directory remains rooted at the admin output tree for every request. Generated files for authenticated non-admin users are written beneath `data/users/<uid>/user`, causing otherwise valid `/api/outputs/...` links to return 404 even though the artifact exists. The bundle provides a concrete reproduction and diagnosis, and the current code still contains the startup-bound mount; there are no competing or prior cross-referenced PRs. The fix must preserve the existing public-output whitelist, private-suffix rejection, traversal protection, and account isolation.

Closes #790

### Related Issues

- Closes #...
- Related to #...

### Module(s) Affected

- [ ] `agents`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `api`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `config`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `core`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `knowledge`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `logging`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `services`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `tools`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `utils`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `web` (Frontend)
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `docs` (Documentation)
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `scripts`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] `tests`
  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] Other: `...`
  Not claimed: the workspace test run did not pass; see the notes above.

### Checklist

  Not claimed: the workspace test run did not pass; see the notes above.
- [ ] My changes do not introduce any new security vulnerabilities.

### Additional Notes

*Add any other context or screenshots about the pull request here.*

### DIFF
--- a/deeptutor/api/main.py
+++ b/deeptutor/api/main.py
@@ -2,7 +2,7 @@ from contextlib import asynccontextmanager
 import logging
 import sys
 
-from fastapi import Depends, FastAPI, HTTPException
+from fastapi import Depends, FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
 
@@ -18,6 +18,13 @@ from deeptutor.services.path_service import get_path_service
 
 ensure_runtime_settings_files()
 export_runtime_settings_to_env(overwrite=True)
+
+# Auth settings are exported above before this module reads their derived
+# constants. Keep this ordering aligned with the deferred router imports below.
+from deeptutor.multi_user.context import user_from_token_payload  # noqa: E402
+from deeptutor.multi_user.paths import get_path_service_for_scope  # noqa: E402
+from deeptutor.services.auth import AUTH_ENABLED, decode_token  # noqa: E402
+
 configure_logging()
 logger = logging.getLogger(__name__)
 
@@ -49,9 +56,28 @@ class SafeOutputStaticFiles(StaticFiles):
         self._path_service = path_service
 
     async def get_response(self, path: str, scope):
-        if not self._path_service.is_public_output_path(path):
+        path_service = self._path_service_for_request(scope)
+        if path_service is None or not path_service.is_public_output_path(path):
             raise HTTPException(status_code=404, detail="Output not found")
-        return await super().get_response(path, scope)
+
+        request_files = StaticFiles(
+            directory=str(path_service.get_public_outputs_root()),
+            check_dir=False,
+            follow_symlink=self.follow_symlink,
+        )
+        return await request_files.get_response(path, scope)
+
+    def _path_service_for_request(self, scope):
+        if not AUTH_ENABLED:
+            return self._path_service
+
+        token = Request(scope).cookies.get("dt_token")
+        payload = decode_token(token) if token else None
+        if payload is None:
+            return None
+
+        user = user_from_token_payload(payload)
+        return get_path_service_for_scope(user.scope)
 
 
 def validate_tool_consistency():
--- a/tests/api/test_output_static_files.py
+++ b/tests/api/test_output_static_files.py
@@ -1,0 +1,152 @@
+"""Request-scoped regression coverage for generated-output downloads."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from pathlib import Path
+
+from fastapi import FastAPI
+from fastapi.testclient import TestClient
+import pytest
+
+from deeptutor.services.auth import TokenPayload
+from deeptutor.services.path_service import PathService
+
+OutputAppFactory = Callable[[dict[str, TokenPayload | None]], tuple[TestClient, PathService, Path]]
+
+
+@pytest.fixture
+def output_app(monkeypatch, tmp_path) -> OutputAppFactory:
+    from deeptutor.api import main as api_main
+    from deeptutor.multi_user import paths as multi_user_paths
+
+    admin_root = tmp_path / "data"
+    users_root = admin_root / "users"
+    admin_service = PathService(workspace_root=admin_root)
+    admin_service.get_public_outputs_root().mkdir(parents=True)
+
+    monkeypatch.setattr(api_main, "AUTH_ENABLED", True)
+    monkeypatch.setattr(multi_user_paths, "USERS_ROOT", users_root)
+    monkeypatch.setattr(multi_user_paths, "_path_services", {})
+
+    def make_app(tokens: dict[str, TokenPayload | None]) -> tuple[TestClient, PathService, Path]:
+        monkeypatch.setattr(api_main, "decode_token", lambda token: tokens.get(token))
+        app = FastAPI()
+        app.mount(
+            "/api/outputs",
+            api_main.SafeOutputStaticFiles(
+                directory=str(admin_service.get_public_outputs_root()),
+                path_service=admin_service,
+            ),
+        )
+        return TestClient(app), admin_service, users_root
+
+    return make_app
+
+
+def _write_output(root: Path, relative_path: str, contents: bytes) -> None:
+    output = root / "user" / relative_path
+    output.parent.mkdir(parents=True, exist_ok=True)
+    output.write_bytes(contents)
+
+
+def test_authenticated_user_downloads_output_from_own_workspace(output_app) -> None:
+    relative_path = "workspace/chat/chat/session-1/code_runs/report.pdf"
+    alice = TokenPayload(username="alice", role="user", user_id="u_alice")
+    client, _admin_service, users_root = output_app({"alice-token": alice})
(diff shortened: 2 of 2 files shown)


mvanhorn:
Looks like this landed upstream independently. 712d413 added deeptutor/api/routers/outputs.py, which serves outputs through a request-scoped path service and fails closed with a 404 when there is no authenticated user, and tests/api/test_output_files.py covers #790 directly.

I compared the two before saying anything: resolve_public_output_path already handles what my tests exercised, traversal outside the outputs root, private suffixes, and non-files. Its docstring also makes a fair point about the shape I used, which validated against one workspace and then rebuilt the path from another root.

Happy to close this out. Let me know if there is anything here worth keeping instead.

pancacake:
Thanks for your contribution — and you diagnosed a genuine multi-user leak here.

`dev` has since replaced the `SafeOutputStaticFiles` mount entirely with a request-scoped router, `deeptutor/api/routers/outputs.py`, which resolves the workspace the same way you proposed:

```python
def _request_path_service() -> PathService:
    user = get_current_user_or_none()
    if user is None:
        raise HTTPException(status_code=404, detail="Output not found")
    return get_path_service_for_scope(user.scope)
```

It goes one step further than this PR in two ways worth noting, since they were the right calls: it fails closed when there is no request user (rather than falling back to the startup-bound admin `PathService`), and it hangs off `Depends(require_auth)` instead of decoding the `dt_token` cookie by hand.

So the bug you found is fixed, just via a different shape — which is also why this PR now conflicts on `deeptutor/api/main.py`: the class it patches no longer exists. Closing as already-fixed. Thanks for catching it.

Issue #779, opened 2026-08-05
[Bug]:write_note/exec tool calls intermittently arrive with empty arguments (missing required mode/command), loop repeats and wastes tokens


hunposhuyang:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

In long chat turns (large tool surface + long history), the LLM intermittently returns a native function call whose arguments is empty ({}) even though the schema marks required fields. Reproduced with both deepseek-chat (api.deepseek.com) and deepseek-v4-flash (OpenAI-compatible gateway). Short, controlled reproductions always succeed — the failure appears under heavy context.
- write_note: in one turn, 5 of 7 calls carried empty arguments (no mode, no notebook_id) → tool returns Unknown mode ''. Use one of: 'append', 'edit'. → the model retries the same empty call 5–7 times, burning the loop budget.
- exec: 3 calls with empty command → ValueError: exec requires a non-empty command. surfaced to the model as a generic "unknown error".
There is no pre-dispatch validation or repair in core/agentic/tool_dispatch.py / runtime/registry/tool_registry.py: empty/partial arguments reach the tool, produce an exception or a hard-to-recover error, and the model repeats the malformed call instead of learning.

### Steps to reproduce

1. Configure an OpenAI-compatible LLM profile (e.g. DeepSeek deepseek-chat).
2. Attach a knowledge base and a notebook so rag, list_notebook, write_note mount; run a multi-turn session with several large RAG results in context.
3. Ask the agent to organize/merge a study note, e.g. "结束今天的学习并整理和修改数据结构笔记本".
4. Observe write_note tool calls whose arguments are empty {}; tool result is Unknown mode ''...; repeated across loop rounds.

### Expected Behavior

- Tool calls are validated against their schema before dispatch; missing required arguments produce a corrective, structured result naming the missing params (so the model can fix in one round) instead of an exception or a generic error.
- A call with empty arguments is not dispatched at all.

### Related Module

Notebook

### Configuration Used

- DeepTutor v1.5.9, Docker single-container (ghcr.io/hkuds/deeptutor:latest)
- Active LLM: deepseek-chat, OpenAI binding, https://api.deepseek.com
- Embedding: local llama.cpp (not involved here)
- main.yaml: chat max_rounds 8

### Logs and screenshots

[...]

Issue #708, opened 2026-07-26
[Bug]:GTP Luna, Terra, and Sol fail to call tools


graysonlee123:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

DeepTutor 1.5.4, OpenAI binding, direct api.openai.com/v1.

GPT 5.4 successfully invokes web_search with the same settings and DuckDuckGo provider configuration.

gpt-5.6-terra, gpt-5.6-sol, and gpt-5.6-luna report that web_search / web_fetch is unavailable or fail to issue tool calls.

Container outbound HTTPS is healthy:
curl -I -L https://example.com succeeds.

Please check tool/function-call support and tool-call response parsing for the gpt-5.6 model family, especially any Chat Completions vs Responses API response-shape mismatch.

### Steps to reproduce

1. Configure an OpenAI model from the gpt-5.6 family
2. Create a fresh chat and ask it "Use the web_search tool to search for DeepTutor GitHub repository. Return the repository URL and the title of the first result. Do not answer from prior knowledge."

### Expected Behavior

Tools should be called and handled properly.

### Related Module

Dashboard

### Additional Information

- DeepTutor Version: 1.5.4
- Operating System: Docker
- Python Version: 3.11.15
- Node.js Version: 22.23.1
- Browser (if applicable): 
- Related Issues:

pancacake:
Thanks — I dug into this and can rule out the obvious suspect, but I can't reproduce it without gpt-5.6 access, so I need one more thing from you.

**What I ruled out.** Tool calling is not gated off for this family:

- `can_use_native_tool_calling(binding="openai", …)` returns `True` — `PROVIDER_CAPABILITIES["openai"]["supports_tools"]` is `True` (`deeptutor/services/llm/capabilities.py:24-32`), so the tool schemas *are* being attached.
- More decisively: **no code path in the repo distinguishes gpt-5.6 from gpt-5.4.** The only occurrence of the string `gpt-5.6` anywhere is an unrelated default constant; every other check is a `"gpt-5"` prefix match that treats both identically. So whatever differs, it isn't a DeepTutor-side gate.

[...]

PR #826, opened 2026-08-12
fix: preserve Responses tool call arguments

### LINKED ISSUE #825: [Bug]: Responses tool arguments are dropped for item_id events or Markdown-fenced content

### Describe the bug

When Codex GPT-5.6 uses the OpenAI Responses API, `write_note` can arrive at dispatch with `{}` even though the model generated all required arguments. The v1.5.10 pre-dispatch guard then reports missing `mode` and `notebook_id`.

Two parser defects combine in the common notebook case:

1. `response.function_call_arguments.delta` and `.done` events may identify the function call with `item_id` only. Both raw SSE and SDK stream consumers currently look up buffers by `call_id`, so those deltas are dropped.
2. After the stream is reconstructed, `parse_json_response()` extracts the first Markdown fence before trying to parse the complete response. A valid tool-argument JSON object whose Markdown `content` string contains a fenced code sample is therefore replaced by the fence body and falls back to `{}`.

This is not caused by a missing skill or Notebook configuration.

### Steps to reproduce

1. Install DeepTutor v1.5.11.
2. Configure Codex GPT-5.6 through the OpenAI Responses provider.
3. Attach a notebook and ask the model to call `write_note` in `edit` mode.
4. Ensure the note `content` contains a fenced Markdown code block.
5. Observe the pre-dispatch rejection:

`write_note was called without its required argument(s): mode, notebook_id`

A deterministic parser reproduction is included in the accompanying PR: one Responses event sequence uses `call_id` on `response.output_item.added`, but only `item_id` on the argument delta/done events; a second test parses valid serialized tool JSON containing a Markdown fence.

### Expected Behavior

Responses argument deltas addressed by either `call_id` or `item_id` should populate the same tool-call buffer. A complete valid JSON object should be parsed before Markdown wrapper extraction. `write_note` should receive the original `mode`, `notebook_id`, and Markdown `content` unchanged.

### Related Module

API/Backend

### Configuration Used

- LLM provider: Codex / OpenAI Responses API
- Model: GPT-5.6
- Tool: `write_note`
- Notebook mode: `edit`
- Logging level used for diagnosis: `DEBUG`

### Logs and screenshots

Sanitized DEBUG evidence (argument values omitted):

```text
args_done call_id=None item_id=fc_... args_len=23175
item_done call_id=call_... item_id=fc_... buffered_len=23175
parsed keys=['content', 'mode', 'note', 'notebook_id', 'record_id', 'title', 'turns_to_include']
Extracted JSON from markdown code block
dispatch model_keys=[] raw_type=str raw_len=23188
Rejected write_note before dispatch: missing required args ['mode', 'notebook_id']
```

After both local fixes, the same end-to-end request reached `write_note` with populated arguments. No credentials, headers, note content, or OAuth data are included here.

### Additional Information

- DeepTutor Version: v1.5.11
- Operating System: macOS 26.5.2
- Python Version: 3.13
- Node.js Version: not relevant
- Browser (if applicable): not relevant
- Related Issues: #779, #708

The accompanying fix targets `dev`, as required by `CONTRIBUTING.md`.

jjjjjjava:
### Description

Fixes two parser failures that can reduce valid OpenAI Responses tool arguments to an empty object:

- Index raw SSE and SDK function-call buffers by both `call_id` and `item_id`, because argument delta/done events may carry only `item_id`.
- Parse a complete valid JSON payload before treating Markdown fences as an outer JSON wrapper. This preserves `write_note.content` when it contains fenced code.
- Add regression coverage for raw SSE, SDK streams, and Markdown-containing serialized tool arguments.

### Related Issues

- Closes #825
- Related to #779
- Related to #708

### Module(s) Affected

- [x] `services`
- [x] `utils`
- [x] `tests`

### Checklist

- [x] Documentation is not required for this behavior-only bug fix.

### Validation

- `tests/services/llm/test_openai_responses_parsing.py`: 2 passed
- `tests/utils/test_json_parser.py`: 35 passed
- Targeted Ruff lint and format checks: passed
- `git diff --check`: passed
- Local end-to-end `write_note` request with fenced Markdown reached tool dispatch with populated arguments.

### Additional Notes

Targeted `pre-commit run --files ...` could not initialize the repository's Prettier hook because the host npm policy rejected its local git dependency with `EALLOWGIT`. No source failure was reported; direct Ruff and pytest checks above passed.

### DIFF
--- a/deeptutor/services/llm/provider_core/openai_responses/parsing.py
+++ b/deeptutor/services/llm/provider_core/openai_responses/parsing.py
@@ -75,31 +75,35 @@ async def consume_sse(
                 call_id = item.get("call_id")
                 if not call_id:
                     continue
-                tool_call_buffers[call_id] = {
+                buffer = {
                     "id": item.get("id") or "fc_0",
                     "name": item.get("name"),
                     "arguments": item.get("arguments") or "",
                 }
+                tool_call_buffers[call_id] = buffer
+                item_id = item.get("id")
+                if item_id:
+                    tool_call_buffers[item_id] = buffer
         elif event_type == "response.output_text.delta":
             delta_text = event.get("delta") or ""
             content += delta_text
             if on_content_delta and delta_text:
                 await on_content_delta(delta_text)
         elif event_type == "response.function_call_arguments.delta":
-            call_id = event.get("call_id")
-            if call_id and call_id in tool_call_buffers:
-                tool_call_buffers[call_id]["arguments"] += event.get("delta") or ""
+            buffer_id = event.get("call_id") or event.get("item_id")
+            if buffer_id and buffer_id in tool_call_buffers:
+                tool_call_buffers[buffer_id]["arguments"] += event.get("delta") or ""
         elif event_type == "response.function_call_arguments.done":
-            call_id = event.get("call_id")
-            if call_id and call_id in tool_call_buffers:
-                tool_call_buffers[call_id]["arguments"] = event.get("arguments") or ""
+            buffer_id = event.get("call_id") or event.get("item_id")
+            if buffer_id and buffer_id in tool_call_buffers:
+                tool_call_buffers[buffer_id]["arguments"] = event.get("arguments") or ""
         elif event_type == "response.output_item.done":
             item = event.get("item") or {}
             if item.get("type") == "function_call":
                 call_id = item.get("call_id")
                 if not call_id:
                     continue
-                buf = tool_call_buffers.get(call_id) or {}
+                buf = tool_call_buffers.get(call_id) or tool_call_buffers.get(item.get("id")) or {}
                 args_raw = buf.get("arguments") or item.get("arguments") or "{}"
                 try:
                     args = json.loads(args_raw)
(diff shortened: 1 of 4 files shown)


pancacake:
Thanks for this — the diagnosis is exactly right, and it landed on `dev` in parallel as c96440d6 ("fix: harden Responses stream parsing (#825)"), which fixes the same two things: correlating argument delta/done events by `item_id` as well as `call_id`, and parsing the whole payload before treating a Markdown fence as a JSON wrapper.

Closing as superseded rather than resolving the conflicts, since every conflicted hunk resolves to dev's side. Your PR did lead somewhere though: reviewing the two side by side surfaced a real bug in the merged version — it aliased buffers under the `fc_0` placeholder, so a call whose `item_id` the provider omitted could be dispatched under a *different* call's tool name and arguments. Your version did not have that. Fixed in 97574b52 with a regression test.

Appreciate the careful write-up.

jjjjjjava:
> Thanks for this — the diagnosis is exactly right, and it landed on `dev` in parallel as [c96440d](https://github.com/HKUDS/DeepTutor/commit/c96440d6aca677fba7e754aec1ae2b085991a0aa) ("fix: harden Responses stream parsing (#825)"), which fixes the same two things: correlating argument delta/done events by `item_id` as well as `call_id`, and parsing the whole payload before treating a Markdown fence as a JSON wrapper.
> 
> Closing as superseded rather than resolving the conflicts, since every conflicted hunk resolves to dev's side. Your PR did lead somewhere though: reviewing the two side by side surfaced a real bug in the merged version — it aliased buffers under the `fc_0` placeholder, so a call whose `item_id` the provider omitted could be dispatched under a _different_ call's tool name and arguments. Your version did not have that. Fixed in [97574b5](https://github.com/HKUDS/DeepTutor/commit/97574b5236e53d4e8894bb101ba2cc2ecf5e1b7e) with a regression test.
> 
> Appreciate the careful write-up.

Thanks for the detailed explanation and for reviewing the two implementations side by side. I’m glad the diagnosis and PR helped confirm the two root causes and also surfaced the fc_0 aliasing issue. I appreciate you fixing it and adding the regression test in 97574b5.

I’ve since rebased onto the latest dev branch and opened #829 as a focused follow-up for separate Responses stream-failure handling gaps and delta-only regression coverage. I’d be happy to make any adjustments based on your feedback.

Thanks again for the review and for maintaining DeepTutor.

PR #830, opened 2026-08-12
fix(mcp): preserve image content for multimodal tools

NoxsMedia:
## Summary

- preserve MCP `ImageContent` as structured `ToolResult.attachments` instead of stringifying its Base64 payload into tool text
- carry image attachments through parallel dispatch, bound to the originating `tool_call_id`
- convert Codex Responses tool images to `function_call_output.output[].type = input_image`
- keep non-Codex providers and existing text-only MCP results on their current message shape

## Problem

`MCPConnectionManager._call_once()` currently stringifies every non-`TextContent` block. For an MCP image result that produces a large textual Pydantic representation containing the Base64 payload. The model therefore receives text, not visual input, while traces and context can be polluted with the encoded image.

The failure was reproduced with PageIndex's `get_document_image`: PageIndex returned a valid JPEG `ImageContent`, but DeepTutor did not send an image part to the model. PageIndex is only the reproducer: the affected input is the standard MCP `ImageContent` type, and the implementation contains no PageIndex-, document-, or language-specific handling.

## Scope and relationship to visual RAG

This PR fixes the runtime transport path for images returned directly by any MCP tool. It does not change document parsing, LlamaIndex ingestion, vector indexing, or retrieval behavior.

#837 addresses a separate ingestion-layer contract for preserving parser-referenced visuals in a LlamaIndex knowledge base. Neither PR depends on the other. Delivering a retrieved LlamaIndex image to a final answer turn would be a separate future change and is not implemented here.

## Implementation

- `ToolResult` gains an `attachments` channel, separate from its LLM-visible text body.
- MCP images are Base64-validated, limited to 10 MiB, restricted to JPEG/PNG/GIF/WebP, and represented in text only by a bounded MIME/size marker.
- Audit metadata records MIME, byte count, and SHA-256, never the Base64 payload.
- Dispatch carries attachments by the exact originating tool call ID.
- The chat loop adds image parts only for `openai_codex`; other provider tool messages remain unchanged.
- The Responses converter emits structured `input_text` / `input_image` blocks in `function_call_output.output`.

### Automated

- `3806 passed, 9 skipped` in a fresh Python 3.11 venv installed exactly from `requirements/server.txt` and `requirements/partners.txt`
- `ruff check .`
- `ruff format --check .`
- repository pre-commit hooks on all changed files, including detect-secrets, Bandit, and mypy
- `python -m compileall -q deeptutor`
- `git diff --check`

The new tests cover:

- text + image MCP results without Base64 leakage into text/metadata
- invalid Base64 and pre-decode image-size rejection
- attachment ownership by `tool_call_id`
- Codex Responses `function_call_output.input_image` conversion
- next-round image injection and unchanged non-Codex message shape

### Real end-to-end regression

Using a fresh session with the PageIndex `1-20` knowledge base and `openai_codex / gpt-5.6-sol`:

1. the model called `mcp_pageindex_get_document_image`
2. a redacted request observer saw `function_call_output.output[].type = input_image`
3. the model read the three transactions, signed amounts, and icon colors from figure 1.1-5
4. the run completed with zero stream errors
5. the redacted report contained no long Base64 runs or data URL payloads

No PageIndex index or production data was changed for this validation.

## Branch coordination

This targets `dev` per `CONTRIBUTING.md` and depends on the Codex vision capability support already merged there in #821. `deeptutor/services/mcp/manager.py` was also changed independently by #813 on `main`; maintainers may need to reconcile that file when the branches converge. This PR intentionally does not copy #813 into `dev` or alter its transport-error scope.

## Design sync

The mechanism is documented here in the PR because the repository's architecture instruction file is protected in the contributor environment. The contract is: tool text remains textual; binary tool evidence stays structured through dispatch and is converted only at a provider boundary.

### DIFF
--- a/deeptutor/agents/chat/agent_loop.py
+++ b/deeptutor/agents/chat/agent_loop.py
@@ -410,6 +410,9 @@ class AgentLoop:
             )
             state.tool_steps += 1
             state.sources.extend(dispatch.sources)
+            self.pipeline._attach_tool_images(
+                dispatch.tool_messages, dispatch.tool_attachments_by_id
+            )
             messages.extend(dispatch.tool_messages)
 
             if dispatch.pause:
--- a/deeptutor/agents/chat/agentic_pipeline.py
+++ b/deeptutor/agents/chat/agentic_pipeline.py
@@ -32,7 +32,7 @@ from deeptutor.core.agentic import (
     dispatch_tool_calls,
 )
 from deeptutor.core.agentic.tool_dispatch import MAX_PARALLEL_TOOL_CALLS
-from deeptutor.core.context import UnifiedContext
+from deeptutor.core.context import Attachment, UnifiedContext
 from deeptutor.core.stream_bus import StreamBus
 from deeptutor.core.tool_protocol import ToolLookup
 from deeptutor.core.trace import (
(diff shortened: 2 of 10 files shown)


NoxsMedia:
Closing in favor of #848, which extracts the reusable provider-neutral bug fix: standard MCP ImageContent is no longer stringified into LLM-visible tool text with its full Base64 payload. The broader attachment/Codex image-transport path in this PR is intentionally not carried forward.

PR #848, opened 2026-08-13
fix(mcp): omit image payloads from tool text

NoxsMedia:
### Description

MCP `ImageContent` blocks currently fall through to `str(block)`, which serializes the full Base64 `data` field into LLM-visible tool text. A single image can therefore consume substantial context and appear in tool traces even though DeepTutor does not currently deliver that block as visual input.

This PR handles standard MCP image blocks explicitly and replaces them with a fixed, bounded marker:

```text
[MCP image omitted]
```

Text blocks and all other MCP content types retain their existing behavior.

### Scope

This is intentionally a safe degradation only. It does **not** add an attachment protocol, provider-specific image transport, PageIndex handling, or visual RAG behavior. Multimodal MCP delivery can be designed separately without retaining the current Base64 leakage.

This supersedes the narrowly reusable part of #830; that older PR also implemented a much broader Codex image-transport path.

### Verification

- `180 passed` — `tests/services/mcp`
- `ruff check` on both changed files
- `ruff format --check` on both changed files
- `python -m compileall` on both changed files
- `git diff --check`
- added-line security scan: no matches

### Design sync

`design-sync: not applicable` — this preserves the current unsupported-image behavior while bounding its textual representation; it does not introduce a new runtime boundary or transport contract.

### DIFF
--- a/deeptutor/services/mcp/manager.py
+++ b/deeptutor/services/mcp/manager.py
@@ -557,6 +557,8 @@ class MCPConnectionManager:
         for block in result.content:
             if isinstance(block, types.TextContent):
                 parts.append(block.text)
+            elif isinstance(block, types.ImageContent):
+                parts.append("[MCP image omitted]")
             else:
                 parts.append(str(block))
         return "\n".join(parts) or "(no output)"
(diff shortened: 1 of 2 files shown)


NoxsMedia:
CI note: all Python 3.11–3.14 test jobs, all import checks, and lint/format pass. The only failing job is the unchanged Web Node baseline: `dev@23ad80d1` already fails the same Web Node workflow, while this PR changes only `deeptutor/services/mcp/manager.py` and its Python MCP test. No unrelated Web change is included.

Issue #641, opened 2026-07-18
[Feature Request]:Add a per-conversation reasoning-effort selector to the chat composer


crasuna:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

Please add a reasoning-effort selector directly to the chat composer, similar to the controls available in ChatGPT and Codex.

At present, there is no visible per-conversation control for reasoning effort in the chat interface. Users should be able to change the reasoning level before sending a message, without restarting the backend or changing a global configuration.

A suitable location would be next to the model or capability selector in the chat composer.

For example:

Selected model · Reasoning: High

The available options could include:

Provider default
None or Minimal
Low
Medium
High
XHigh
Max

Only the values supported by the selected model and provider should be shown.

The selected value should:

Apply to the next submitted message.
Remain active in the current conversation until changed.
Not affect other conversations.
Take effect without restarting the backend.
Be clearly visible before a message is sent.
Be shown in the Activity panel or runtime details after the request is completed.

A possible configuration precedence would be:

Current conversation setting
LLM profile default
Global environment configuration
Provider or model default

This would preserve existing configuration methods while allowing task-level control from the chat interface.

For models or providers that do not support reasoning effort, the selector should be hidden or disabled. The application should also avoid sending unsupported parameter values.

### Related Module

Frontend/Web

### Use Case

Different requests require different balances between response quality, latency, and resource usage.

A lightweight request may only require a low or medium reasoning level, while a more complex request may benefit from a higher level. Allowing users to adjust this setting from the chat composer would avoid applying the same reasoning level to every request and would eliminate the need to restart the backend when changing it.

### Additional Context

Suggested acceptance criteria:

Reasoning effort can be changed directly from the chat interface.
The change takes effect on the next message without restarting DeepTutor.
The selected value persists within the current conversation.
Different conversations can use different values.
Only supported values are sent to the selected provider.
The effective value is visible in the Activity panel or runtime details.
Existing global configuration remains available as a fallback.

ImproveRei0:
不至于在每一个会话中都单独设置，但是我认为在模型配置界面增加**关于OpenAI模型的reasoning_effort配置选项**是有必要的
似乎在调用openai api时，如果这个值未被传入，那么模型能力将默认为medium，
[...]

PR #958, opened 2026-08-22
feat: add per-conversation reasoning-effort override (#641)

ademicho123:
Description: Adds an optional per-conversation reasoning_effort override to LLMSelection, the existing model-pin primitive that already flows from a turn's payload (or saved session preferences) through to the real LLM call. When set, it takes precedence over the profile/model's own default; when unset, existing resolution is unchanged. The effective value is also exposed in the turn metadata that already surfaces llm_model/llm_provider (Activity panel data). The chat-composer selector UI itself is a separate frontend change, not included here.

Related Issues: Related to #641

Module(s) Affected: services, api

Additional Notes: 9 new tests added (tests/services/model_selection/test_llm_selection.py, tests/services/config/test_provider_runtime.py); full suite run clean against a fresh dev checkout.


ademicho123:
@pancacake I believe this is fix now.

Issue #967, opened 2026-08-23
feat(reading): support faithful EPUB rendering in the reading pane


evan188199-tech:
## Problem

The existing Reading pane imports EPUB files as roughly 2,800-character plain-text sections. That preserves assistant grounding, but loses the publication's XHTML/CSS, illustrations, fonts, spine order, pagination, and durable EPUB positions.

## Proposed focused stack

Extend the existing `deeptutor.reading` / `ReaderPane` foundation rather than adding a parallel reading workspace:

1. Add a backward-compatible EPUB locator contract (`render_mode`, `unit_refs`, CFI anchors, and persisted positions) while keeping legacy manifests in text mode.
2. Render authenticated original EPUB bytes inside `ReaderPane` with epub.js, map spine hrefs back to numeric locators, and preserve assistant jumps and annotations.

The viewer should keep scripted content disabled, safely handle external links, support LTR/RTL keyboard and touch pagination, and work on narrow iPhone/iPad layouts.

## Compatibility and scope

- PDF and plain-text reading remain unchanged.
- Existing EPUB manifests remain in text mode; no destructive automatic migration.
- This foundation does **not** include translation, dictionaries, read-aloud, quizzes, Kids permissions/rewards, or a separate reader workspace.
- Follow-up bilingual reading work can build on this contract; related discussion: #860.
- This proposal does not depend on #719, #905, or #917.

## Acceptance

- Original EPUB styles and images render in the existing Reading pane.
- Numeric locator ↔ spine href mapping remains compatible with assistant citations.
- CFI progress and annotations survive refresh, with unique-quote fallback for legacy annotations.
- Corrupt EPUBs show an error instead of an endless loading state.
- Desktop Chromium and mobile WebKit coverage accompanies backend compatibility/security tests.

evan188199-tech:
Completed upstream by #968 (source-faithful EPUB locator/position contract) and #969 (faithful EPUB rendering in ReaderPane), including original XHTML/CSS/images, locator mapping, CFI progress/annotations, and desktop/mobile verification.

PR #968, opened 2026-08-23
feat(reading): add source-faithful EPUB contract

evan188199-tech:
## Summary

Adds a backward-compatible, source-faithful EPUB contract to the existing `deeptutor.reading` engine. This is the backend foundation for faithful EPUB rendering in `ReaderPane`; it does not introduce a second reading module or a new page.

Part of #967; the stacked viewer PR completes the user-facing portion.

## What changed

- Parse EPUB reading units in spine order through the shared secure ZIP/EPUB parser.
- Map nested EPUB3 navigation and EPUB2 NCX entries to numeric locators.
- Preserve the original EPUB and expose canonical `source_href` values through `unit_refs`.
- Add `render_mode`, optional annotation `source_anchor`, and GET/PUT reading position APIs.
- Keep `has_raw_view` PDF-only so this PR remains safe with the old frontend.
- Default legacy manifests to text mode.
- Upgrade a re-uploaded legacy EPUB atomically only when it has no annotations; otherwise return an explicit 409 conflict.
- Exclude XHTML `<head>` content and scripts from extracted reading text.

## Compatibility and security

- Existing PDF, text, and legacy EPUB reading flows remain available.
- Locator, percentage, and CFI length are validated server-side.
- Raw files and positions stay inside the current user's workspace.
- Existing archive traversal, compression-ratio, entry-count, and expanded-size protections are reused.

## Verification

- `pytest tests/reading tests/utils/test_document_extractor.py -q` — **202 passed**
- Ruff on Reading, router, and affected tests — **PASS**
- `git diff --check` — **PASS**

Covered cases include spine order, nested nav/NCX, images, RTL/no-nav EPUBs, corrupt archives and ZIP bombs, href mapping, position/CFI validation, workspace isolation, legacy manifests, and re-upload conflicts.

## Out of scope

No viewer UI, translation, dictionary, TTS, quiz, Kids flow, rewards, or Focus Check.


evan188199-tech:
Review hardening: legacy EPUB re-upgrades no longer copy a stale text-reader position. The old locator space could have more units than the source-faithful spine, so GET position could return an out-of-range locator after upgrade. The viewport now resets to chapter 1 while annotation protection remains unchanged.

Verification on 9daea4d0:
- pytest tests/reading tests/utils/test_document_extractor.py -q — 203 passed
- Ruff check/format on deeptutor/reading/store.py and tests/reading/test_engine.py — passed
- git diff --check — passed

Remaining review note: the two-rename backup installation narrows the partial-ingest window but is not crash-atomic between renames; a process death in that gap can leave the material temporarily missing with a backup directory. This should either be documented as recoverable-on-next-ingest or followed up with startup recovery before merge.

Issue #380, opened 2026-04-24
[Feature Request]: Add a Learning Experience Plugin SDK + first-party “Knowledge Garden” / “Concept Companion” learning layer


yepyhun:
YES! I asked ai help to write my ideas into ticket , but these are my ideas and tips that i'm missing / would be interesting.
This is the reason I left a comment before I left this project cause it's abandoned...
Cause these suggestions are very important (for me at least) and it's a ton of work and development phase is too slow to do these in a meaningful phase.
Why would I use deeptutor if notebooklm is useable and first class implemented bugless ect ....
You should offer something more than the bare minimum...

## Do you need to file a feature request?

* [x] I have searched the existing feature requests and I do not believe this exact feature request is already filed.
* [x] I believe this is a legitimate feature request, not just a question or bug.
* [x] This request is focused on long-term learning usefulness, not on deployment, SaaS, or installation convenience.

---

## Feature Request Description

DeepTutor already has many of the hard ingredients of a serious AI learning platform: persistent memory, TutorBots, Question Bank, Book Engine, Skills, RAG-backed knowledge bases, notebooks, multi-mode chat, visualization, deep solving, and research workflows.

What still feels missing is a dedicated **Learning Experience Layer**: a plugin system that allows developers and educators to build motivational, adaptive, long-term study mechanics on top of real learner behavior.

This is not a request for shallow gamification such as XP, coins, streaks, badges, or generic leaderboards.

Instead, the proposal is to add a **Learning Experience Plugin SDK** that can react to study events, update learner-state models, render small interactive UI surfaces, and create meaningful study rituals. The goal is to make DeepTutor feel less like a set of AI tools and more like a living academic companion that evolves with the learner over months or years.

As a first-party example plugin, I propose something like:

> **Knowledge Garden**
> A living visual map where concepts behave like plants/ecosystems. Concepts grow when the learner recalls, explains, applies, and connects them. Weak concepts wilt, misconceptions appear as weeds, prerequisite gaps show up as shallow roots, and interdisciplinary connections become cross-pollination paths.

Or alternatively:

> **Concept Companions**
> A mature, non-childish version of “catch/train/battle” learning. Instead of Pokémon-like creatures, each concept or topic becomes an evolving companion/entity whose behavior reflects the learner’s actual understanding. A concept companion does not level up from time spent. It develops through demonstrated recall, explanation quality, transfer tasks, and correction of misconceptions.

[...]

PR #970, opened 2026-08-23
feat(reading): add safe extension protocol and empty toolbar slot

evan188199-tech:
## Summary

- add a versioned `deeptutor.reading_extensions` entry-point contract with no built-in extensions
- expose authenticated, server-verified action contexts for material locators and selections
- render only validated `card`, `quiz`, `feedback`, and `browser_speech` results; never plugin HTML or JavaScript
- hide the Reader extension toolbar completely when no extension package is installed
- isolate discovery and action failures so optional packages cannot break reading

## Security boundary

The server reloads the requested locator from `ReadingStore`; browser-supplied visible text is not accepted. A selection is forwarded only if it occurs in that stored unit. Result payloads have a 64 KB ceiling plus quiz/speech shape limits.

## Scope

This PR intentionally includes **no** Learn, Test, TTS, Translate, Vocabulary, reward, or model implementation. Those belong in optional packages. It is independent of the EPUB and web-source PRs.

## Verification

- `python -m pytest tests/reading -q` — 161 passed
- focused Ruff — passed
- `npm run test:node` — 589 passed
- `npx tsc --noEmit` — passed
- touched-file ESLint with `--max-warnings=0` — passed
- `npm run i18n:parity` — passed
- `npm run build` — passed, 59 routes

Relates to #380.

### DIFF
--- a/READING_EXTENSIONS.md
+++ b/READING_EXTENSIONS.md
@@ -48,6 +48,10 @@ class ExampleExtension:
   `feedback`, or `browser_speech`.
 - Results have a 64 KB serialized ceiling. Quiz and speech payloads receive
   additional shape and length validation.
+- Units larger than the protocol's 60,000-character context ceiling are
+  rejected with a client error instead of invoking an extension.
+- Actions have a 30-second execution timeout and return the standard
+  recoverable unavailability response when exceeded.
 - Result data is rendered as React text. Extensions cannot send JavaScript or
   raw HTML to the Reader.
 - Discovery and execution failures are isolated. A broken optional package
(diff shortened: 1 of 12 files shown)


evan188199-tech:
Review hardening on fd1d1385: extension actions now handle two runtime edges explicitly. A valid but oversized PDF/EPUB unit (over the 60,000-character protocol context limit) returns 422 instead of an uncaught validation error, and hanging actions return the recoverable 503 response after a 30-second timeout. The protocol documentation now states both bounds.

Verification:
- pytest tests/reading -q — 163 passed
- cd web && npm run test:node — 589 passed
- Ruff check/format on the touched Python files — passed
- git diff --check — passed

One caveat: Python cannot forcibly terminate a synchronous extension that ignores cancellation, but the request now recovers at timeout and the worker remains an installed-package trust boundary rather than browser-controlled code.

Issue #971, opened 2026-08-23
feat(reading): finish selector re-anchoring for structured revisions


evan188199-tech:
## Status update

The original selector foundation is now upstream:

- #968/#969 added source-faithful EPUB rendering.
- #972 added the backward-compatible W3C TextQuoteSelector / TextPositionSelector contract.
- #974 added the Recogito-based reflowing text renderer and legacy-quote fallback.

This issue now tracks only the remaining structured-content and revision cases.

## Remaining scope

1. Verify selector creation, restoration, navigation, and annotation clicks when web snapshots render as rich Markdown rather than flattened plain text (#980).
2. When web snapshot revisions are introduced, use quote prefix/suffix and position context to migrate only reliable matches.
3. Preserve ambiguous or unresolved annotations for explicit review instead of painting the wrong passage or silently dropping data.

## Acceptance criteria

- Highlights remain visible in structured Markdown without flattening headings, links, tables, code, or bilingual disclosure blocks.
- Repeated quotes resolve through bounded context; ambiguity never selects the wrong passage.
- Old quote, PDF rectangle, and EPUB CFI annotations remain readable.
- Revision migration migrates reliable selectors, records unresolved/ambiguous rows, and preserves the previous annotation for audit.
- Backend migration and frontend rich-content behavior are covered by focused tests.

## Non-goals

- No new selector data model, annotation renderer, PDF geometry path, or EPUB CFI path.
- No Learn, Test, read-aloud, Translate, or Vocabulary implementation in core.

PR #972, opened 2026-08-23
feat(reading): persist W3C text selectors

evan188199-tech:
## Summary

Add an optional W3C text-selector contract to Immersive Reading annotations so reflowing renderers can persist both quoted text and its position without changing the existing PDF rectangle flow.

Closes the first backend-contract phase of #971.

## Changes

- Add bounded `TextQuoteSelector` and `TextPositionSelector` models.
- Round-trip selectors through annotation JSON storage and the Reading API.
- Keep `quote` backward compatible; a quote selector can populate an omitted legacy `quote`.
- Reject mismatched `quote` / `TextQuoteSelector.exact` values and invalid position ranges.

## Compatibility and migration

- Existing annotation JSON does not need migration: missing `selectors` parses as an empty tuple.
- Existing API clients can continue sending `quote`, `rects`, or both.
- PDF rendering and export behavior is unchanged.

## Security and limits

- The API accepts at most two selectors.
- Quote text is capped at 2,000 characters; prefix and suffix context are capped at 128 characters each.
- Positions must be non-negative and `end > start`.
- Unknown selector shapes are rejected by the API and ignored when reading legacy/corrupt stored JSON.

## Tests

- `159 passed`: `.venv/bin/python -m pytest tests/reading -q`
- Ruff lint: passed on all six touched files.
- Ruff format check: passed on all six touched files.
- `git diff --check`: passed.

## Out of scope

This PR deliberately does not add a frontend annotation library, revision migration, translation, TTS, quizzes, vocabulary, or other learning actions. A reflowing-text renderer and revision re-anchoring can follow as separate reviewable PRs under #971.


evan188199-tech:
Follow-up hardening pushed in 6332d339: TextQuoteSelector.exact must occur in the server-stored reading unit; TextPositionSelector cannot extend beyond the unit or span more than 2,000 characters; API and store layers enforce the span; legacy selector parsing now preserves the beginning of an overlong prefix. Verification: `uv run --extra dev pytest tests/reading -q` (163 passed), `ruff check` on touched files (pass), `ruff format --check` (pass), and `git diff --check` (pass).

evan188199-tech:
Full CI on head 6332d339 is green: Repository Hygiene, Import Checks (Python 3.11–3.14), Lint and Format, Python Tests (3.11–3.14), Test Summary, and Web Node Tests all passed.

Issue #846, opened 2026-08-13
[Feature Request]: Support DeepSeek Responses API and native server-side web search


bitons:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

DeepSeek now officially supports the Responses API for both `deepseek-v4-flash` and `deepseek-v4-pro`, including the server-side `web_search` tool.

Please add first-class DeepSeek Responses API support to DeepTutor. In DeepTutor 1.5.11, the Responses routing logic does not select DeepSeek, and the main agentic loop (Chat/Solve/Question/Research) calls Chat Completions directly, bypassing the OpenAI-compatible provider path.

Expected behavior:

- DeepSeek V4 Flash and Pro can use the Responses API.
- DeepTutor maps its `web_search` tool to the native Responses tool schema: `{ "type": "web_search" }`.
- Agentic Chat, Solve, Question, and Research use the same supported provider path instead of bypassing it.
- Other providers and configured local search backends keep their current behavior.

### Related Module

API/Backend

### Use Case

When a user selects a DeepSeek V4 model and enables web search, DeepTutor currently receives a function call named `web_search` and executes the configured local backend (for example DuckDuckGo or SearXNG).

This prevents users from using DeepSeek's native server-side search even though their model and API support it. Native routing would simplify configuration, avoid unnecessary local search calls, and use the provider capability the user selected.

### Additional Context

Official DeepSeek documentation: https://api-docs.deepseek.com/guides/responses_api

Local validation:

- A direct DeepSeek Responses request with `tools: [{ "type": "web_search" }]` returned a `web_search_call`.
- With the current DeepTutor agentic path, the same task invoked the local DuckDuckGo integration.
- After locally adapting the agentic client to route DeepSeek through the Responses provider, native search completed successfully and no local DuckDuckGo result artifact was created.

Likely areas involved:

- `services/llm/provider_core/openai_compat_provider.py` (Responses API selection)
- Responses tool conversion (map DeepTutor's `web_search` function to the native tool)
- `core/agentic/client.py` (the direct Chat Completions path currently bypasses provider routing)

It would also be helpful to add a regression test confirming that DeepSeek agentic requests use the native Responses web-search tool rather than a local search backend.

ningkaikok:
I'd like to work on the **native server-side `web_search` tool mapping** half of this request.

**Scope I'm taking:**
[...]

PR #978, opened 2026-08-24
feat(llm): route OpenAI-compatible agent calls through the Responses API

### LINKED ISSUE #727: [Docs]: Feature request: Support Responses API for custom OpenAI-compatible providers
### Before filing

- [x] I have searched the existing docs issues and this report is not already filed.
- [x] This is about the docs at https://deeptutor.info — not a product bug.

### Page URL

https://deeptutor.info/zh-cn/get-started/providers/

### What kind of issue?

Other

## Background

The custom OpenAI API profile currently appears to use the Chat Completions protocol by default.

My provider supports the model `gpt-5.6-luna`, but this model does not support Chat Completions. The API returns:

```json
{
  "message": "模型 gpt-5.6-luna 不支持 chat completions 协议",
  "type": "codexzh_api_error",
  "code": "protocol_not_supported"
}

### What did you expect?

{
  "binding": "custom",
  "base_url": "https://api.example.com/v1",
  "protocol": "responses"
}

### Environment (optional)

_No response_

qian-123456:
## Summary

All agentic pipelines (chat loop, labeled steps, research, question, explore) stream through a raw `AsyncOpenAI` client built by `core/agentic/client.py`, which always targets `{base}/chat/completions`. The existing Responses API support in `services/llm/provider_core` never applies on this path because it is gated to direct `api.openai.com` connections (#882 covered the direct-OpenAI case only).

For users behind OpenAI-compatible gateways/relays this means prompt caching is effectively unreachable: every agent round replays the full conversation at full input price. #727 reports the hard version of this — some gateways reject Chat Completions entirely for newer models (`protocol_not_supported`).

## What this PR does

Adds an opt-out **Responses bridge** at the shared client factory (`core/agentic/client.py`), so `chat.completions.create(...)` is transparently served by `{base}/responses`:

- **Request side**: chat kwargs are converted to Responses bodies using the existing converters in `services/llm/provider_core/openai_responses/` (`convert_messages`, `convert_tools`). Multiple system messages are merged into `instructions` (the converter otherwise keeps only the last one). Chat-only kwargs such as `stream_options` are dropped.
- **Response side**: Responses SDK events and objects are re-shaped into chat-completions chunks (`delta.content`, `delta.reasoning_content`, `delta.tool_calls`, `finish_reason`, `usage` incl. `prompt_tokens_details.cached_tokens`). Tool-call arguments are buffered and emitted exactly once per call so consumers that accumulate deltas by concatenation stay correct. No consumer code changes required.
- **Failure handling**: any bridge error retries the original request via Chat Completions for that call, and feeds a per-(base URL, model) circuit breaker — after 2 failures the gateway is left alone for 5 minutes, then probed again. Gateways without `/responses` therefore degrade gracefully instead of breaking.
- **Activation policy** via `DEEPTUTOR_USE_RESPONSES_API`: `on` / `off` / unset-or-anything (auto). Auto bridges cloud OpenAI-compatible bindings and skips local inference servers (Ollama, vLLM, LM Studio, …).

## Testing

- New unit tests (`tests/core/agentic/test_responses_bridge.py`, 14 cases): activation policy, system-message merging + kwarg dropping, tool schema conversion, stream event shaping (single tool-call emission, finish reason, usage-only final frame with cached tokens), non-stream shape, fallback + breaker trip/expiry.
- Regression: `tests/core` 187 passed; `tests/services/llm` 171 passed, 1 pre-existing skip.
- End-to-end smoke against a local mock gateway exposing both endpoints: streaming served by `/v1/responses`; non-stream conversion; a 404 on `/responses` falls back mid-request to chat completions; breaker stops probing after two failures.

Fixes #727
Relates to #846 (DeepSeek Responses support would slot into the same seam), builds on #882.

Happy to flip the default from auto to opt-in only if maintainers prefer a more conservative rollout.

### DIFF
--- a/deeptutor/core/agentic/client.py
+++ b/deeptutor/core/agentic/client.py
@@ -27,6 +27,7 @@ from deeptutor.services.llm import get_token_limit_kwargs, supports_tools
 from deeptutor.services.llm.reasoning_params import (
     build_openai_compatible_reasoning_kwargs,
 )
+from deeptutor.core.agentic.responses_bridge import wrap_responses_bridge
 from deeptutor.services.provider_registry import find_by_name
 
 # Providers that don't reliably support OpenAI function-calling. The loop
@@ -99,12 +100,13 @@ def _build_openai_client(config: LLMClientConfig, *, disable_ssl_verify: bool) -
             http_client=http_client,
             default_headers=default_headers,
         )
-    return AsyncOpenAI(
+    client = AsyncOpenAI(
         api_key=config.api_key or "sk-no-key-required",
         base_url=config.base_url or None,
         http_client=http_client,
         default_headers=default_headers,
     )
+    return wrap_responses_bridge(client, config)
 
 
 async def _close_client(client: Any) -> None:
--- a/deeptutor/core/agentic/responses_bridge.py
+++ b/deeptutor/core/agentic/responses_bridge.py
@@ -115,6 +115,54 @@ def _record_bridge_success(key: str) -> None:
         _bridge_tripped_at.pop(key, None)
 
 
+#: Statuses that mean "this attempt failed, but the endpoint itself is
+#: fine" — relay channel exhaustion (new-api 503), rate limits, upstream
+#: hiccups. Falling back for the request is right; tripping the breaker
+#: would only convert a one-request blip into a five-minute all-chat
+#: window (zero prompt-cache hits), so these never count toward it.
+_TRANSIENT_STATUS_CODES = frozenset({429, 500, 502, 503, 504})
+
+
+def _failure_status_code(exc: BaseException) -> int | None:
+    status = getattr(exc, "status_code", None)
+    if status is None:
+        status = getattr(getattr(exc, "response", None), "status_code", None)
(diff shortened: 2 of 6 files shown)


pancacake:
Thanks for your comments! Closing because `dev` already has this.

The Responses API path lives in `deeptutor/services/llm/provider_core/openai_responses/` (`converters.py`, `parsing.py`, with tests), and `OpenAICompatProvider._should_use_responses_api` routes agent calls through it, including a circuit breaker (`_responses_circuit_allows`) and native `web_search` tool mapping. #1075 extended that to `deepseek-v4-pro` earlier today.

The bigger practical issue: this branch forks at `v1.5.12` and is **265 commits** behind `dev` — the diff against current `dev` is 5039 files with ~106k deletions, so it can no longer be merged without destroying a large amount of unrelated work. It was also still targeting `main`; I retargeted it to `dev` before checking, which is how the drift surfaced.

If there is a specific behaviour in your bridge that the shipped path still gets wrong, a fresh small PR off current `dev` would be very welcome.

PR #1022, opened 2026-08-27
fix(mastery): normalize objective module variants

### LINKED ISSUE #1019: Mastery Paths save empty for most LLMs — _parse_modules only reads name, models emit title

### Describe the bug

**Mastery Paths save empty for most LLMs** because `_parse_modules` only reads the `name` field, but the models emit `title`.

The tool schema (`deeptutor/capabilities/mastery/tools.py`, `_parse_modules`) requires:

```json
modules: [ { "name": "...", "knowledge_points": [ { "name": "...", "type": "..." } ] } ]
```

But the docs and system prompt describe the plan as made of **"objectives"** typed as **"concept or a procedure"**, and never show the exact JSON shape. As a result, LLMs (tested: gemma4, qwen3.5, gpt-oss) emit structurally-broken variants that `_parse_modules` rejects:

1. `{ "objectives": [ { "id", "title", "description", "type" } ] }` — wrong top-level key, no `name`
2. `modules: [ { "knowledge_points": ["concept_framework", ...], "name": "..." } ]` — KPs as bare strings
3. `modules: [ { "knowledge_points": [ { "id": "..." } ], "name": "..." } ]` — KPs as dicts with only `id`
4. `modules: [ { "description", "id", "type" } ]` — flat KP list, no `knowledge_points` key

Because `_parse_modules` only reads `raw_kp.get("name")`, every knowledge point is skipped → the path saves empty and never appears in Learning Space. The failure is **silent** (no user-facing error), which is why it's hard to catch.

### Steps to reproduce

1. Open Chat, select the **Mastery Path** capability.
2. Ask it to build a Mastery Path from your materials (attach a KB or PDF).
3. Watch the Activity panel — `mastery_build` is called but rejected (`"No valid modules..."`), retried, rejected again.
4. Open Learning Space → Mastery Path — nothing there (or an empty path).

### Expected behavior

The Mastery Path should save with the modules/knowledge points the model designed, regardless of whether the model emits `name`/`title`/`id` or uses `knowledge_points`/`objectives`.

### Actual behavior

The path is created empty and never appears in Learning Space. The failure is silent and model-dependent.

### Environment

- DeepTutor version: 1.5.11 (Docker image `ghcr.io/hkuds/deeptutor:latest`)
- Deployment: Docker, single container
- LLM provider: Ollama Cloud (OpenAI-compatible `https://ollama.com/v1`)
- Models tested: `gemma4:31b`, `qwen3.5:397b`, `gpt-oss:120b`

### Suggested fix

Make `_parse_modules` a lenient salvage parser that:
- accepts `modules` **or** `objectives` at the top level,
- derives a KP name from `name` → `title` → `id` → slugified `description`,
- accepts knowledge points as dicts, bare strings, or a flat list.

Also consider aligning the docs/system-prompt terminology ("objectives" vs "knowledge_points") with the actual tool schema so models emit the correct shape in the first place.

evan188199-tech:
## Summary
- Normalize common `mastery_build` model variants: wrapped `modules`/`objectives` payloads, `name`/`title`/`id`/`description` fallbacks, bare-string objectives, and flat objective lists.
- Keep generated storage IDs server-controlled and preserve the existing type fallback.
- Tighten EN/ZH prompts around the canonical `modules=[{name, knowledge_points: [{name, type}]}]` shape.
- Reject explicitly empty module trees and skip empty modules without converting them into fabricated objectives.

Fixes #1019

## Test plan
- [x] `ruff format --check deeptutor/capabilities/mastery/tools.py deeptutor/learning/tests/test_mastery_tools.py`
- [x] `ruff check deeptutor/capabilities/mastery/tools.py deeptutor/learning/tests/test_mastery_tools.py`
- [x] `pytest tests/capabilities/test_mastery_capability.py deeptutor/learning/tests/test_mastery_tools.py -q` (79 passed)


pancacake:
Thanks for your comments!

Closing as superseded by #1079, which is merged. Your diagnosis and the two-sided treatment were right — state the shape the tool wants, and parse permissively for the models that ignore it — and #1079 keeps that direction.

Two differences worth noting:

- The normalization is a single pass that reduces any input to `[(module name, knowledge points)]`, rather than the three-branch `entries` construction here (where `_parse_knowledge_point` was computed and then discarded, used only as an emptiness test).
- It adds a name-precedence order (`name` → `title` → `label` → `objective` → `topic` → `description` → `id`) so variant 4 from #1019, which carries only `description`/`id`/`type`, is readable too, and guards the identifier-humanizing step so CJK names are never reshaped.

The prompt change here was dropped deliberately: the tool schema already declares the shape with `required` fields, so #1079 puts it in the failure message instead, where it reaches a retrying model exactly when it is needed rather than costing tokens every turn.

PR #1048, opened 2026-08-27
fix(llm): replay DeepSeek reasoning across tool calls

Wubian111:
### Description

DeepSeek thinking-mode requests made through the Responses API fail on the round after a local tool call with:

```
The `reasoning_text` in the thinking mode must be passed back to the API.
```

The knowledge-base request itself succeeds, but DeepTutor reconstructs the assistant turn from visible text and function calls. That drops the provider's original `reasoning` output item, so the next request sends the RAG result without the required reasoning state and DeepSeek rejects it.

This patch keeps the fix scoped to Responses turns that actually contain a reasoning item:

- preserve the completed Responses output items for a reasoning/tool-call turn;
- carry those items through the chat agent loop;
- replay the exact items before the function-call output on the next request;
- preserve exact Responses `call_id` values only for replay messages.

Non-reasoning Responses turns retain the existing web-search metadata behavior. Ordinary Responses conversion and Chat Completions tool-id normalization retain their existing paths.

DeepSeek documents that thinking-mode tool calls must return the complete reasoning content in subsequent requests: https://api-docs.deepseek.com/zh-cn/guides/thinking_mode/

### Related Issues

- No linked issue; reproduces the reported DeepSeek thinking-mode RAG failure.

### Module(s) Affected

- [x] `agents`
- [ ] `api`
- [ ] `config`
- [ ] `core`
- [ ] `knowledge`
- [ ] `logging`
- [x] `services`
- [ ] `tools`
- [ ] `utils`
- [ ] `web` (Frontend)
- [ ] `docs` (Documentation)
- [ ] `scripts`
- [x] `tests`
- [ ] Other: `...`

### Testing

- `pytest -q tests/agents/chat/test_agent_loop.py tests/services/llm/test_openai_compat_reasoning_content.py tests/services/llm/test_openai_responses_converters.py tests/services/llm/test_openai_responses_parsing.py` — 82 passed.
- `ruff check` and `ruff format --check` on all eight changed files — passed.
- Ubuntu 24.04 `pre-commit run --files <all eight changed files>` — all applicable hooks passed, including hygiene, Ruff, detect-secrets, Bandit, and MyPy.
- A full Windows test run reached 5,397 passed and 31 skipped; 72 failures were isolated to existing Windows/POSIX assumptions, optional integrations, and sparse-checkout packaging fixtures. None were in the touched test modules.

### Checklist

- [x] I have updated the documentation (if necessary).

### Additional Notes

`pre-commit run --all-files` was run in a clean Ubuntu 24.04 clone. Every hook except the pinned Prettier hook passed. Prettier rewrote 37 pre-existing Web files that are unchanged between this branch and `dev`; including those unrelated formatting changes would violate the scope of this bug fix. After restoring those baseline files, the complete applicable hook set passes for every file changed by this PR. This PR intentionally contains no Web diff.


simongu:
### Independent confirmation + interim mitigation suggestion

We independently hit the exact error this PR fixes on a `dev`-based workspace (v1.6.2):

```
Error: {'message': 'The reasoning_text in the thinking mode must be passed back to the API.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}
```

Repro: `deepseek-v4-flash` via the `deepseek` binding, chat agent loop with `web_search` mounted → the request routes through the Responses API (`_uses_native_web_search`), the first round succeeds with native web search, and the continuation round after a local tool call (e.g. `rag`) 400s. Same root cause as analyzed here: `convert_messages()` drops the provider's `reasoning` output item.

Two additions this PR might want to consider:

1. **Residual gap: poisoned histories.** Replay depends on `_responses_output_items` being present on the prior assistant message. Turns persisted before this fix (or reused after a provider/model switch) have no stored items, so the 400 still fires post-merge — and since the error text contains none of the current markers, `_should_fallback_from_responses_error()` returns `False` and users get a hard failure instead of the existing Chat Completions degradation. Suggest adding `reasoning_text` / `passed back` / `thinking mode` markers to `_should_fallback_from_responses_error` as defense-in-depth: the Chat Completions path already sends `thinking: disabled` for `deepseek-v4-flash` (#1058), so the fallback is safe and preserves the turn.

2. We verified the marker approach against the live error payload (`status_code=400`, body containing the message above) — with those markers added, `_should_fallback_from_responses_error` returns `True`.

Happy to open this as a separate stopgap PR if you prefer to keep #1048 scoped.

Related: #1175 (general Responses reasoning-item replay) and #1058 / #1067 (prior Chat Completions reasoning persistence).

Wubian111:
Thanks @simongu for independently confirming the root cause and for opening #1195.

After re-reviewing this branch against the current `dev`, the provider response-state persistence and Responses replay path is now covered by #1067 and subsequent backend changes. The remaining poisoned-history fallback is addressed separately by #1195, while #1175 tracks the broader Responses reasoning-item replay work.

Keeping #1048 open would now duplicate that work and retain conflicts without an independent minimal production delta, so I am closing it as superseded. The original analysis here remains valid and helped identify the missing replay path.

Thanks again for the concrete reproduction and the defense-in-depth follow-up.

PR #1078, opened 2026-08-29
fix(runtime): skip psutil process-tree walk on HarmonyOS (/proc/1/stat)

### LINKED ISSUE #1076: [Bug]: GET /api/v1/system/memory 500 (PermissionError: /proc/1/stat) on HarmonyOS/OpenHarmony

### Describe the bug
On **HarmonyOS / OpenHarmony** (not standard Linux), unprivileged processes cannot read most `/proc/<pid>/stat` entries. The system memory endpoint crashes with an unhandled `PermissionError` and returns HTTP 500.

The memory probe uses psutil's process-tree walk when available: `deeptutor/runtime/memory_probe.py` -> `_scan_psutil()` -> `root.children(recursive=True)`. psutil's `children()`/`_ppid_map()` iterates **every** `/proc/<pid>/stat`, and reading the init process `/proc/1/stat` raises a raw `PermissionError`. Because the probe only catches `psutil.Error` (and `PermissionError` is not a subclass of it), the exception escapes to the HTTP layer.

```
PermissionError: [Errno 13] Permission denied: '/proc/1/stat'
  memory_probe.py:143 _scan_psutil -> root.children(recursive=True)
  psutil/__init__.py:972 children -> _ppid_map()
  psutil/_pslinux.py:1694 ppid_map -> open_binary(procfs_path/pid/stat)
GET /api/v1/system/memory -> 500
```

### Steps to reproduce
1. Run DeepTutor backend on HarmonyOS / OpenHarmony (restricted unprivileged /proc access).
2. Request `GET /api/v1/system/memory`.
3. Observe HTTP 500 with `PermissionError: /proc/1/stat`.

### Expected Behavior
`GET /api/v1/system/memory` returns 200 with a valid memory snapshot instead of crashing. The probe should gracefully skip unreadable `/proc` entries; the codebase already ships a psutil-free fallback walker `_scan_proc()` that reads each `/proc/<pid>/stat` defensively and skips unreadable pids.

### Related Module
API/Backend

### Configuration Used
- HarmonyOS / OpenHarmony (aarch64, musl) with restricted unprivileged `/proc` access
- DeepTutor 1.6.0

### Logs and screenshots
See the traceback in the "Describe the bug" section.

### Additional Information
- DeepTutor Version: v1.6.0
- Operating System: HarmonyOS / OpenHarmony (aarch64, musl)
- Python Version: 3.12
- Node.js Version: n/a
- Browser (if applicable): n/a
- Related Issues: none found (searched `proc/1/stat`, `memory_probe`, HarmonyOS/OpenHarmony)
- Proposed fix: in `_scan_psutil()`, widen the `except` to `(psutil.Error, PermissionError, OSError)` and fall back to the existing psutil-free `_scan_proc()`; verified locally that `capture()` returns a valid snapshot.

wotomchen:
## Summary
On **HarmonyOS / OpenHarmony**, unprivileged processes cannot read most `/proc/<pid>/stat` entries. `GET /api/v1/system/memory` therefore crashed with HTTP 500 because `psutil`'s process-tree walk (`children()`/`_ppid_map()`) iterates every `/proc/<pid>/stat` and raised `PermissionError` on the unreadable init process `/proc/1/stat`, which is not a `psutil.Error` subclass and so escaped the probe's handler.

## Fix
Detect HarmonyOS via `os.uname().sysname == "HarmonyOS"` and, on that platform, use the existing psutil-free `_scan_proc()` walker — which reads each `/proc/<pid>/stat` defensively and skips unreadable entries — instead of psutil's process-tree enumeration. Behaviour on Linux/Docker is **unchanged** (the new path only runs when HarmonyOS is detected).

## Verification
On HarmonyOS (aarch64, musl) before the fix:
```
>>> root = psutil.Process(os.getpid()); root.children(recursive=True)
PermissionError: [Errno 13] Permission denied: '/proc/1/stat'
```
After the fix:
```
>>> memory_probe.capture()
MemorySnapshot(total_rss_bytes=43905024, partial=True, ...)  # 200, no crash
```

Closes #1076.

### DIFF
--- a/deeptutor/runtime/memory_probe.py
+++ b/deeptutor/runtime/memory_probe.py
@@ -79,6 +79,20 @@ class MemorySnapshot:
         return self.total_rss_bytes / self.limit_bytes
 
 
+def _is_harmonyos() -> bool:
+    """Detect HarmonyOS/OpenHarmony, where unprivileged ``/proc`` access is restricted.
+
+    On HarmonyOS a non-root process cannot read most ``/proc/<pid>/stat``
+    entries, so psutil's ``children()``/``_ppid_map()`` raises
+    ``PermissionError`` on the unreadable init process. The psutil-free
+    ``_scan_proc()`` walker skips such entries instead, so it is used there.
+    """
+    try:
+        return os.uname().sysname == "HarmonyOS"
+    except (AttributeError, OSError):
+        return False
+
+
 def _load_psutil() -> Any | None:
     """Import psutil lazily; it is a soft dependency (see ``_scan_proc``)."""
     try:
@@ -136,6 +150,14 @@ def _classify(pid: int, name: str, cmdline: str, root_pid: int | None = None) ->
 
 def _scan_psutil(psutil: Any) -> tuple[list[ProcessMemory], bool]:
     """Walk the supervisor's tree with psutil (the cross-platform path)."""
+    if _is_harmonyos():
+        # HarmonyOS/OpenHarmony hide most ``/proc/<pid>/stat`` entries from
+        # unprivileged processes, so psutil's ``children()``/``_ppid_map()``
+        # raises ``PermissionError`` on the unreadable init process. Walk
+        # the tree with the psutil-free ``/proc`` reader instead, which skips
+        # unreadable entries defensively.
+        return _scan_proc()
+
     root_pid = _supervisor_pid(psutil.pid_exists)
     partial = root_pid is None
     try:


pancacake:
Thanks for your comments!

Closing in favour of #1077, which is now merged — but the diagnosis in this PR and in #1076 is what made the fix possible. Your traceback pinned it exactly: `children()` → `_ppid_map()` → `open_binary(/proc/1/stat)`, raising a raw `PermissionError` that is not a `psutil.Error` subclass and so escaped the probe's handler.

We went with the behaviour-based fix rather than the platform check for two reasons:

1. It catches the raw `PermissionError`/`OSError` wherever it comes from, so hardened containers, gVisor/seccomp sandboxes and any other restricted `/proc` get the same fallback — not only hosts whose `uname().sysname` happens to read `HarmonyOS`.
2. It also covers two more spots that raise the same escaping errors: `proc.cmdline()` and `_scan_psutil_self`.

It is also what you asked for in #1076's Expected Behavior — "the probe should gracefully skip unreadable /proc entries".

If you get a chance, a confirmation that `GET /api/v1/system/memory` now returns 200 on your HarmonyOS aarch64/musl device would be genuinely useful — yours is the only real device this has been seen on.

PR #1134, opened 2026-08-31
fix(video-learning): retry unavailable Invidious captions (#1133)

### LINKED ISSUE #1133: fix(video-learning): retry unavailable Invidious captions
### Problem

Invidious playback can resolve successfully while the initial caption request fails or returns no usable cues. Playback remains available, but transcript-guided learning is unavailable and there is no in-place recovery path.

### Expected behavior

An authenticated learner can retry caption retrieval for the current Invidious material without recreating the material, resetting playback/progress, or falling back to YouTube. A failed retry must preserve the stored material.

### Proposed scope

- Add an authenticated, material-scoped transcript refresh endpoint.
- Retrieve metadata and captions only from the configured Invidious instance.
- Refresh the subtitle track URL so the existing HTML5 player reloads captions without remounting.
- Surface a retry control only for unavailable captions on Invidious playback.

### Reproduction

1. Configure an Invidious provider and open a YouTube learning video.
2. Cause the initial Invidious captions request to fail while video metadata/streams still resolve.
3. Observe that playback works but captions and Explain here remain unavailable.
4. Restore the provider; there is currently no in-page way to retry captions.

evan188199-tech:
### Description

Adds an explicit recovery path when Invidious streams resolve but captions are temporarily unavailable. The refresh is material-scoped and Invidious-only: it preserves the stored material, playback descriptor, and learning progress on success, and leaves stored state untouched when the provider request fails. The returned subtitle URL carries a revision so the existing HTML5 video track reloads without remounting the player.

### Related Issues

- Closes #1133

### Module(s) Affected

- [x] `api`
- [x] `services`
- [x] `web` (Frontend)
- [x] `tests`

### Verification

- [x] `ruff check` and `ruff format --check` on changed Python files
- [x] Focused video-learning tests: 35 passed (1 existing environment-sensitive test deselected)
- [x] `npm run test:node` (774 passed)
- [x] `npm run lint` (0 errors; 57 pre-existing warnings)
- [x] `npx tsc --noEmit`
- [x] `npm run i18n:check`
- [ ] Playwright browser audit — blocked locally by unstable isolated Next/Chromium environment; the regression scenario is included in `web/tests/e2e/video-learning.audit.ts`.

### Additional Notes

The full focused Python selection has one existing failure unrelated to this PR: its route inspection expects flattened FastAPI routes, while the installed FastAPI exposes `_IncludedRouter` wrappers.


evan188199-tech:
Retargeted this PR from dev to main; the head is already based on the current main commit. /retest

PR #1152, opened 2026-08-31
fix(runtime): ignore Windows SIGINT for deeptutor start (#1147)

### LINKED ISSUE #1147: [Windows] Opening the browser sends CTRL_C_EVENT (SIGINT) that kills the launcher; conflicts with Ctrl+C
## Summary

On **Windows**, running `deeptutor start` in the foreground and then opening the frontend in a browser makes the launcher receive a stream of `CTRL_C_EVENT` (SIGINT), which shuts down the whole app. The same signal is also what a real `Ctrl+C` produces, so the launcher cannot distinguish "browser opened" from "user pressed Ctrl+C". This makes it essentially impossible to (a) let the browser be opened/used without killing the server, and (b) stop it with `Ctrl+C`.

## Environment

- Windows 10/11, Windows Terminal
- Conda env (Python 3.13), deeptutor 1.6.2 (PyPI install)
- Model provider configured (opencode.ai/zen/go/v1) + Tavily search

## Steps to reproduce

1. `conda activate deeptutor`
2. `deeptutor start`
3. The terminal prints `请在浏览器中打开 http://localhost:3782` and the backend + frontend come up (FastAPI on 8001, Next.js on 3782).
4. Open the frontend in a browser (type the URL, or let it open).
5. The terminal immediately logs `收到 SIGINT 信号,正在关闭 ...` (repeated several times) and shuts down the whole app (frontend + backend).

## Root-cause analysis

- The **launcher** (`deeptutor.exe`) runs in the **foreground console process group**. The backend (`uvicorn`) and frontend (`Next.js`/node) children are spawned with `CREATE_NEW_PROCESS_GROUP` (see `deeptutor/runtime/launcher.py` `_spawn`), so **they are isolated** from the console signal — only the **launcher** receives it.
- When the browser opens (or takes focus), Windows delivers `CTRL_C_EVENT` → the launcher's SIGINT handler (`_install_signal_handlers`) calls `request_shutdown`, which terminates everything.
- On Windows, the browser-open signal and a deliberate `Ctrl+C` are **both `CTRL_C_EVENT` (SIGINT)** — indistinguishable at the OS level. Opening the browser actually produces a *stream* of these events (the terminal shows 4+ `收到 SIGINT` messages in a row), so any "require N signals" or time-window heuristic is unreliable.
- A separate console window (`CREATE_NEW_CONSOLE`) did **not** reliably isolate it either — the signal still reached the launcher.

## What I tried / workarounds (for reference)

- **Detached launch** (`Start-Process -WindowStyle Hidden` / `CREATE_NO_WINDOW`) + launcher **ignores SIGINT** (`signal.SIG_IGN` on SIGINT): the browser can never kill it; but `Ctrl+C` is disabled (stop via `stop.bat` / taskkill).
- **Launcher ignores SIGINT, stop via `Ctrl+Break` (SIGBREAK)**: browser-safe + keyboard stop, but stop is `Ctrl+Break`, not `Ctrl+C`.
- I could not find a way to have both "browser-open never kills it" and "Ctrl+C stops it", because they are the same signal on Windows.

## Suggested fixes for the maintainers

1. **Auto-open the browser via `webbrowser.open(url)` / `ShellExecute`** (as Streamlit/Gradio do) and print the URL — this avoids the Windows Terminal hyperlink `Ctrl+Click` which sends a spurious `CTRL_C_EVENT`; and the launcher should not rely on the user "clicking a link" in the terminal.
2. **Make the launcher resilient to spurious console signals on Windows**: ignore `SIGINT`/`CTRL_C_EVENT` (browser-open) and support a dedicated stop (e.g. `Ctrl+Break` / `SIGBREAK`, a `.stop` file, or a `deeptutor stop` subcommand). It is not possible to keep `Ctrl+C` working while being immune to browser-open `SIGINT`, because they are the same signal.
3. **Consider not placing the launcher in the foreground interactive console** during startup (e.g. accept a `--daemon`/`--background` flag), or normalize the whole server tree into a single process group/console so Windows console events are handled consistently.

This is a real UX problem for Windows users on a normal `deeptutor start` workflow. Happy to help test any fix.

ZQR1101:
## Summary
- On Windows, opening the frontend (or Ctrl+Clicking the terminal URL) delivers `CTRL_C_EVENT`/`SIGINT` to the foreground launcher — the same signal as a deliberate Ctrl+C — so `deeptutor start` shuts down (#1147).
- Ignore `SIGINT` on Windows and stop via `Ctrl+Break` (`SIGBREAK`); keep Unix Ctrl+C behavior unchanged.
- Best-effort `webbrowser.open` after ready so users need not Ctrl+Click the printed URL; update the stop hint copy.

## Test plan
- [x] `pytest tests/runtime/test_launcher.py::test_windows_signal_handlers_ignore_sigint tests/runtime/test_launcher.py::test_non_windows_signal_handlers_keep_sigint tests/runtime/test_launcher.py::test_stop_hint_uses_ctrl_break_on_windows tests/runtime/test_launcher.py::test_open_frontend_in_browser_is_best_effort -q` (4 passed)
- [ ] On Windows: `deeptutor start`, open/focus the browser — server stays up
- [ ] On Windows: `Ctrl+Break` stops the launcher; hint text mentions Ctrl+Break
- [ ] On Linux/macOS: Ctrl+C still stops as before

Fixes #1147


KryptonGao:
I came across another approach that may be worth considering instead of globally ignoring SIGINT on Windows.

The idea is to keep the normal foreground deeptutor start behavior unchanged, including Ctrl+C, and add a separate detached mode for Windows.

The implementation I saw ( @wjunhere, #1147, In the group chat on WeChat ) used:

* CREATE_NO_WINDOW for the managed backend/frontend processes
* DETACHED_PROCESS for the background launcher
* a new deeptutor start --detach command
* a corresponding deeptutor stop command
* a runtime launcher.json file to record the launcher/backend/frontend PIDs and ports
* OpenProcess + GetExitCodeProcess for Windows PID liveness checks instead of os.kill(pid, 0)
* a separate log file for detached startup

So the workflows would be:

deeptutor start
→ normal foreground mode, with Ctrl+C still working

deeptutor start --detach
→ Windows-safe detached mode

deeptutor stop
→ safely stops only the registered DeepTutor processes

I think this may solve the Windows console-signal issue without changing the expected Ctrl+C behavior for every Windows user.

A simple .bat (#1154) launcher could also call the detached mode for users who prefer a one-click startup experience.

pancacake:
感谢提交。#1147 已由 dev 的 `7134ea6a` 以更完整方案解决：修复 Windows 上破坏性的 PID 存活探测、自动打开浏览器，并提供可正常停止的 detached 模式；前台模式继续保留 Ctrl+C，因此不采用对所有 Windows 前台启动全局忽略 SIGINT 的方案。该 PR 作为已被替代关闭。

Issue #1181, opened 2026-09-02
[Bug]: Gemini tool-calling loop fails with 400 "missing thought_signature" on multi-round function calls


garcher123:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

  
  
Environment:  
- Component: deeptutor.agents.chat.agent_loop (mastery capability, but affects any tool-using chat turn)  
- Provider/binding: gemini (openai_compat backend, https://generativelanguage.googleapis.com/v1beta/openai/)  
  
Summary:  
Any chat/mastery turn where Gemini makes a tool call (e.g. `mastery_status`) fails on  
the following round with:  
  
  Error code: 400 - [{'error': {'code': 400, 'message': 'Function call is missing a  
  thought_signature in functionCall parts. This is required for tools to work  
  correctly ... Additional data, function call `default_api:mastery_status`,  
  position N.', 'status': 'INVALID_ARGUMENT'}}]  
  
The agent loop's error-salvage path logs "agent loop round failed after N round(s);  
forcing finish" and returns a forced/partial answer instead of continuing the  
tool-calling loop.  
  
Root cause:  
Gemini's "thinking" models require a `thought_signature` returned alongside each  
`functionCall` part to be echoed back in the corresponding function-call message on  
the following request. DeepTutor routes Gemini as a generic `openai_compat` backend  
and builds assistant tool-call messages via `assistant_message_with_tool_calls()`  
(deeptutor/core/agentic/messages.py), which only serializes `id`, `type`, and  
`function.{name,arguments}` — there is no `thought_signature` field captured or  
replayed anywhere in the codebase. Because of this, every follow-up round after a  
Gemini tool call is rejected by the API.  
  
Impact:  
- Any Gemini-backed session using tool calling (mastery paths, and potentially other  
  tool-enabled chat capabilities) degrades to a forced/partial answer after the first  
  tool call, rather than completing the intended multi-round tool loop.  
  
Steps to reproduce:  
1. Configure DeepTutor with the Gemini provider/binding for chat.  
2. Start a mastery-path chat session.  
3. Trigger a tool call (e.g. mastery_status is called first-turn automatically).  
4. Observe the WARNING in backend logs: "agent loop round failed after 1 round(s);  
   forcing finish" with the thought_signature 400 error, and the resulting reply  
[...]

PR #1090, opened 2026-08-29
fix(web): remap KaTeX selections to LaTeX for Little Tutor

ZQR1101:
## Summary
- Selecting rendered KaTeX in chat produced Unicode glyphs / broken superscripts via `Selection#toString()`, so Little Tutor failed with `Selected text was not found in the authoritative source message` (#1089).
- Remap intersecting `.katex` / `.katex-display` nodes to their `application/x-tex` annotation, wrapped as `$...$` / `$$...$$` to match stored Markdown.
- Apply the same remap on copy from the messages pane.

## Test plan
- [x] `npm run test:node -- selection-tutor` (664 passed)
- [ ] In chat, select a display math formula and open Little Tutor — selected text should be `$$...$$` TeX, not rendered glyphs; grounding should succeed
- [ ] Copy a KaTeX formula from chat — clipboard should contain Markdown math delimiters
- [ ] Selecting non-math prose still behaves as before

Fixes #1089

PR #1185, opened 2026-09-02
fix(agentic): harden Gemini thought_signature capture

ZQR1101:
﻿## Summary
- Tip already ships the base #1181 fix (replay `extra_content` on assistant tool calls)
- This PR hardens capture so signatures are not dropped when deltas arrive as **plain dicts** or only under SDK **`model_extra`**
- `assistant_message_with_tool_calls` can also rebuild `extra_content` from a bare `thought_signature` string
- Include tip route-budget lazy chunks so Web Node Tests stay green (same approach as #1216 / #1219; Related #1215)

Related to #1181

## Test plan
- [x] `pytest tests/core/agentic/test_tool_call_stream.py tests/core/test_agentic_messages.py -q` (14 passed)
- [x] `ruff check` on touched agentic files
- [ ] CI Web Node Tests green after rebase


ZQR1101:
﻿Closing this as **superseded** after today's v1.6.5 reconcile (dev → `42fab3cf`).

Analysis:
- The core fix for #1181 (preserve Gemini `extra_content.google.thought_signature` and replay it on the next assistant tool-call message) is already implemented upstream, authored by the maintainers themselves — see `deeptutor/runtime/agentic/messages.py` and `tool_call_stream.py` since `93df3d48` / `02ba5679` (both files carry the `#1181` comment). The reconciled `dev` keeps that implementation and only adds unrelated `thinking_blocks` / `part_at` work on top.
- What this PR adds on top is defensive-only: reading the extension out of `model_extra` or plain-dict deltas and a bare `thought_signature` fallback. All three `ToolCallAccumulator` call sites (`agents/loop/agent_loop.py`, `capabilities/explore_context/explorer.py`, `runtime/agentic/labeled_step.py`) feed SDK delta objects, so those shapes never occur on the current code path.

Same precedent as #1090 / #1189 (upstream already carries an equivalent fix → close rather than carry a redundant branch). Happy to reopen if any call site starts passing raw dict deltas and the hardening becomes load-bearing.

ZQR1101:
Superseded by upstream-maintained thought_signature replay; see analysis above.

PR #1186, opened 2026-09-02
fix(update): detect source checkouts before enabling auto-update

dongyu-will:
## Summary
- Detect source checkouts before enabling application self-update.
- Keep source installations on their normal Git/developer update path.
- Remove unrelated reader, WebSocket-formatting, CI, and route-test commits that were previously stacked on this branch.
- Rebased onto current `dev`.

## Verification
- `pytest tests/services/test_app_update.py -q` — 11 passed
- `ruff check` and `ruff format --check` on touched files — passed.

### DIFF
--- a/deeptutor/services/app_update.py
+++ b/deeptutor/services/app_update.py
@@ -166,6 +166,15 @@ def _distribution_direct_url() -> dict[str, Any] | None:
     return payload if isinstance(payload, dict) else {}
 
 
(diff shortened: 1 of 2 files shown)


dongyu-will:
Hi @pancacake, when you have a chance, could you please review this fix? It prevents source checkouts from being treated as auto-updateable installations, and all CI checks are passing. Thank you!

PR #1189, opened 2026-09-02
fix(mastery): restore Save to Notebook and turn navigator

### LINKED ISSUE #1182: [Feature Request]:[Mastery Path] Navigation bar and "Save to Notebook" missing from Mastery Path conversations (v1.6.3)

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

## Summary

After upgrading to v1.6.3, the **navigation bar** and **"Save to Notebook"** features are no longer available in Mastery Path conversations. Both features are still present in regular chat conversations.

## Context

- **DeepTutor version:** v1.6.3
- **OS:** Windows

In Mastery Path mode, the left panel shows a **progress outline** (learning waypoints), but it only serves as a progress indicator — it does **not** allow navigating to a specific conversation turn. Previously, this outline was displayed above the input box for quick progress reference.

## The problem

Mastery Path conversations are **purposefully created** to learn a specific skill or subject. They contain far more knowledge worth saving than casual chat conversations. Yet the two tools most useful for organizing that knowledge — the navigation bar and "Save to Notebook" — were removed from Mastery Path, while they remain in regular chat.

This seems counterintuitive: the mode that produces the most valuable content has the fewest tools to preserve it.

## Expected behavior

- The **navigation bar** should allow jumping to a specific turn within a Mastery Path conversation.
- The **"Save to Notebook"** button should be available in Mastery Path conversations, just as it is in regular chat.

## Question

Was this removal intentional, or a regression introduced by the v1.6.3 front- and back-end refactor? If intentional, could these features be restored for Mastery Path, given that its conversations are the most worth preserving?

## Workaround

Currently exporting conversations as Markdown manually, but this is tedious compared to the previous in-conversation "Save to Notebook" workflow.

### Related Module

Dashboard

### Use Case

_No response_

### Additional Context

_No response_

ZQR1101:
## Summary
- Restore the chat **turn navigator** (left-gutter jump rail) on Mastery Study, matching `ChatWorkspace`: `buildChatOutline` + `TurnNavigator` over a positioned scrollport with `data-chat-column`.
- Restore the header **Save to Notebook** action and `SaveToNotebookModal` so mastery sessions can be filed like regular chat (metadata tags `source: mastery_path`).
- Keep the existing progress `StudyOutline` rail; it answers “where am I on the map,” while the turn navigator answers “jump to a prior question.”

## Related
- Fixes #1182

## Test plan
- [x] `cd web && npm run test:node -- mastery-study-notebook-nav`
- [ ] Open `/mastery/{pathId}/sessions/{sessionId}` with a multi-turn transcript — left gutter ticks jump to turns; Alt+↑/↓ still work
- [ ] Header bookmark opens Save to Notebook with selectable messages; save lands in a notebook
- [ ] Empty / loading study states leave Save disabled and hide the turn rail

### DIFF
--- a/web/components/space/learning/MasteryStudy.tsx
+++ b/web/components/space/learning/MasteryStudy.tsx
@@ -5,6 +5,7 @@ import { browserStorage } from "@/shared/storage";
 import Link from "next/link";
 import {
   ArrowLeft,
+  BookmarkPlus,
   Compass,
   Flag,
   Loader2,
@@ -13,7 +14,8 @@ import {
   PanelRight,
   Sparkles,
 } from "lucide-react";
-import { useCallback, useEffect, useRef, useState } from "react";
+import dynamic from "next/dynamic";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useTranslation } from "react-i18next";
 
 import { ChatMessageList } from "@/features/chat/messages";
@@ -22,6 +24,7 @@ import { buildSessionActivity } from "@/components/chat/home/SessionActivityPane
 import SessionViewerPanel, {
   type SessionViewerPanelHandle,
 } from "@/components/chat/home/SessionViewerPanel";
+import { TurnNavigator } from "@/components/chat/home/TurnNavigator";
 import { useChatStateAdapter } from "@/features/chat/ChatStateAdapter";
 import { TurnStatusBar } from "@/features/chat/components/turn";
 import { turnViewState } from "@/features/chat/model/turn-state";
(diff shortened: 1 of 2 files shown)


ZQR1101:
Closing as superseded.

Issue #1182 is already CLOSED (fixed in v1.6.5 per maintainer). Current `dev` `MasteryStudy.tsx` already wires:
- `TurnNavigator` + `buildChatOutline` / `scrollToChatTurn`
- header Save to Notebook + `buildConversationNotebookSave` / `SaveToNotebookModal`

Rebase would just re-apply an older inline variant on top of the tip helpers, so this PR no longer adds value.

ZQR1101:
Superseded by tip MasteryStudy (issue #1182 closed for v1.6.5).

Issue #740, opened 2026-07-31
[Feature Request]:  Introduce stateful persistent learning(state referencing Matt Pocock's /teach skill design)


woyeyao:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

学习本身是有状态(state)的持久化工程，而非一次性无状态问答交互。所以建议参考Matt Pocock 开源的 /teach 技能。
可以先结合学习目标与当前水平，制定具体计划，然后每次学习后记录当前状态，方便下次接着学。这样更适合可迭代的长期学习路线。
该需求和 DeepTutor 现有的 Memory等可以互补兼容，不需要推翻现有架构，仅需新增一层学习进程的结构化状态管理层，能大幅强化产品长期深度学习的核心竞争力。
Looking forward to your evaluation on this feature proposal, thanks for maintaining this excellent open-source tutoring project.

### Related Module

Other

PR #1226, opened 2026-09-04
feat(learning): soft journal for multi-session teaching

ZQR1101:
## Summary

**Stage 1 — soft learning journal.** Adds a journal (mission + last-session handoff + ADR-style learning records) under the per-user workspace (`learning_journal/journal.json`), inspired by Matt Pocock's `/teach` skill, and exposes always-on chat tools `learning_status` / `learning_update` (`set_mission` | `note_session` | `add_record`). It complements Memory (preferences) and Mastery Path (scored curriculum) and replaces neither.

**Stage 2 — actually resume.** Stage 1 was pull-only: the model had to think to call `learning_status`, so opening a new conversation and saying "continue where we stopped" still got a blank slate — the one thing #740 asks for. Now the turn executor snapshots the journal *before* the loop runs and the chat assembler mounts it as its own system block beside `memory`:

- Only mission + last-session handoff are injected; **records stay behind `learning_status`**, so a turn's overhead never grows with journal history.
- The snapshot is taken once at turn start, which keeps the system prompt **byte-stable for the whole turn** (one shared cache prefix) even after the model writes to the journal mid-turn.
- An empty journal injects **nothing** — no feature advertisement in every unrelated turn. Selection-tutoring turns stay isolated from it, exactly as they stay isolated from global memory.
- The block earns its own context-budget segment (`learning_journal`), with en/zh labels, instead of silently folding into the `capability` line in the composer readout.
- `learning_status`'s description and prompt hints now point at the record history rather than telling the model to re-read state it already has.

**Rebased for the v1.6.5 reconcile (2026-09-06).** `dev` was reconciled with `main` (`42fab3cf`, v1.6.5 release); this branch now sits on top of it. The incidental route-budget web lazy-chunks that the earlier tip needed were **dropped** — the release sync brought its own web structure, so the shims are obsolete (5 commits, 24 files, +1065/−2). The journal feature itself is unchanged: storage, the always-on tools, the turn-start snapshot and its `learning_journal` prompt block all merged onto the new module layout (`prompt_blocks`/`context_budget` moved to `agents/loop/` upstream).

## Related

- Part of #740 (journal store + cross-session resume; not a full `/teach` HTML lesson pipeline).

## Test plan

**Covered by the automated suite** — `pytest tests/services/learning_journal tests/tools/test_learning_journal_tools.py tests/agents/chat/test_learning_journal_injection.py tests/agents/chat/test_context_budget.py tests/api/test_unified_ws_turn_runtime.py::test_selection_tutoring_never_mounts_the_learning_journal` (**33 passed**). A wider run of `tests/api/test_unified_ws_turn_runtime.py tests/agents/chat` shows **no new failures against pristine v1.6.5 `dev`** (identical 2 env failures there: Windows / `test_message_build` no-LLM-config assertions unrelated to this feature):

- [x] `ruff check .` and `ruff format --check .` (0.16.0, as CI runs) clean
- [x] A new conversation already knows the mission and next focus, without the model being asked to look — `test_journal_snapshot_mounts_as_its_own_block`, plus a turn-runtime test asserting the snapshot reaches `UnifiedContext`
- [x] Empty journal → no block and no extra tokens — `test_injection_snapshot_is_empty_for_a_new_learner` (store returns `""`) and `test_empty_journal_mounts_nothing` (assembler mounts nothing)
- [x] Selected-text tutoring stays out of the journal — `test_selection_tutoring_never_mounts_the_learning_journal` asserts the store is never even consulted, and fails when the executor's `if not selection_tutor_context` guard is removed
- [x] Records never ride along — `test_injection_snapshot_leaves_records_out`, so per-turn overhead stays flat as the journal grows

**Verified live against a running app + `deepseek-chat` (2026-09-07)** — full stack built from this branch, driven through the real browser UI:

- [x] Cross-session resume with nothing restated: a brand-new conversation opened with only "let's continue" answered from the journal ("according to the learning journal, we built the frequency-domain = ingredient-list intuition…") and resumed exactly at the recorded `next_focus` (orthogonality), without the model being asked to look anything up
- [x] Composer context readout shows **Learning journal** (en) / **学习日志** (zh) as its own line — 379 tokens for the handoff snapshot, beside Messages / System tools / Skills
- [x] Selected-text tutoring stays out of *global* memory — captured the Little Tutor's `start_turn` frame on the wire: `selection_tutor_context` fully populated (selected text + source message), and the `memory_references` key is **absent from the payload entirely**, while normal chat turns carry it
- [x] Tool boundaries hold: "remember I prefer everyday examples" → `write_memory` (not the journal); "what's my Mastery Path progress on section X" → the tutor consulted `learning_status`, then **declined to answer** ("that's scored curriculum, not in the learning journal") and pointed at the Mastery UI. Note `mastery_status` is not mounted in plain chat, so only this direction of the boundary is observable here
- [ ] One honest caveat: `learning_update` firing *unprompted* is model- and moment-dependent — in the first session it wrote only after being asked; in the second it wrote by itself at a milestone. A missed write makes the inherited handoff stale rather than wrong, which is exactly what the auto-handoff follow-up in the stack is for.

> **CI note (2026-09-07):** `Web Node Tests` is green again: once the settings-bootstrap stub let the readiness panel mount, the blocker notice and the matrix row both render the same unreachable-parser text, so the audit's `panel.getByText(...)` was rejected by Playwright strict mode (resolved to 2 elements). The assertion is now scoped to the matrix row (e523701d) and was verified locally red → green against a production server.
> The branch also carries the `#1274` patches (brand-icon chunk off the app shell + the readiness-audit stub) so the base Web debt does not mask this PR, plus a `style:` commit ruff-formatting the two `deeptutor/skills/builtin/*/SKILL.md` files the v1.6.5 reconcile shipped unformatted (repo-wide `Lint and Format` fails on the base without it).
> **Remaining red is base debt only**: `Python Tests (3.12)` — `deeptutor/learning/tests/test_mastery_mode.py` cannot find `zh/mastery_loop.yaml` on CI; pristine v1.6.5 `dev` fails identically (unrelated to this PR). `Test Summary` merely rolls that up.


ZQR1101:
Stage 3 for #740 sits on top of this branch's head, in case it is useful context while reviewing: https://github.com/ZQR1101/DeepTutor/pull/1

What it adds: `learning_update` is a voluntary tool call, so when the model skips it, the handoff that stage 2 carries into a new conversation is silently stale — which is worse than not carrying it, because the tutor continues from the wrong place. Stage 3 consolidates once after DONE (the same metadata window the session title already uses), and only if the tutor did not write the handoff itself during the turn.

I deliberately kept it out of this PR: one extra LLM call per turn is a decision that deserves its own yes/no, and #1226 is already green as scoped.

Two things about its status:

- It is a fork-internal PR (base = this branch), so it reports no checks — Actions are not enabled on my fork. CI will run when I open it against `dev` once this one lands.
- Everything is fail-open there (no mission, no transcript on both sides, timeout, unparsable output → nothing is written), so a learner who never opened the journal pays nothing at all.

No urgency on it — raising #1226 first is fine.

Issue #781, opened 2026-08-05
[Feature Request]: Support per-user OpenAI Codex OAuth in multi-user deployments


NoxsMedia:
### Do you need to file a feature request?

- [x] I have searched the existing issues and pull requests, and this request is not already filed.
- [x] I believe this is a legitimate feature request rather than a support question.

### Feature Request Description

Support **per-user OpenAI Codex OAuth login** in authenticated multi-user deployments.

The goal is not to share an administrator's ChatGPT/Codex subscription through grants. Each ordinary user should be able to connect and use their own Codex account, while credentials, model visibility, logout, refresh, and inference remain isolated to that user.

This would also align the implementation with the current README, which says that an owner-bound Codex profile is not shareable through grants and that “each account signs in for itself.”

### Related Module

API/Backend and Frontend/Web (multi-user model access and Settings → LLM).

### Use Case

In a family, classroom, or team deployment:

1. An administrator enables authentication and creates an ordinary user.
2. The administrator may grant shared API-key-backed models to that user.
3. The ordinary user signs in to their own ChatGPT/Codex account from their own session.
4. Their personal Codex models appear only in their model selector.
5. Their OAuth tokens and catalog remain inaccessible to the administrator UI and all other users.

Today, this flow cannot be completed. Codex profiles are correctly marked `owner_bound` and withheld from grants, but the OAuth Settings endpoints require an administrator. As a result, ordinary users can neither receive an admin Codex profile nor sign in to their own account.

### Current Behavior and Source Evidence

Verified against `v1.5.9` / commit `37c3db6df7e886aee4f61c97ec5e618b8ab379e8`:

- `deeptutor/api/routers/settings.py` applies `_require_settings_admin()` to Codex OAuth start, status, cancel, refresh, and logout.
- `deeptutor/multi_user/router.py` excludes owner-bound profiles from assignable admin resources.
- `deeptutor/multi_user/model_access.py` drops owner-bound profiles from ordinary-user grant resolution.
- `deeptutor/services/codex_auth/service.py` and `deeptutor/multi_user/paths.py` already provide useful foundations:
  - credentials are stored under `data/system/user-secrets/<owner>/...`;
  - OAuth service instances are keyed by the owner's secret root;
  - the callback broker routes callbacks by OAuth `state` across service instances.

The missing piece is a safe personal-model access path for ordinary users.

[...]

PR #1253, opened 2026-09-05
fix(codex): recover Docker OAuth callbacks in settings

### LINKED ISSUE #1252: Codex OAuth has no in-page recovery when localhost callback cannot reach Docker
### Problem

When DeepTutor runs in Docker and the Web UI is opened through localhost, CodexOAuthCard treats it as native local execution and opens OAuth directly. The browser returns to localhost:1455/auth/callback, but that loopback listener is inside the container. The standard deployment publishes only Web/API ports, so the user sees a failed callback page and sign-in remains waiting.

### Existing work

PR #743 added an opt-in temporary Compose port bridge. This report concerns the remaining in-product recovery gap: users of the default container configuration see no recovery path in the local Web UI. The remote SSH guidance does not appear on localhost either.

### Proposed fix / acceptance

- Offer a callback-URL completion form for any waiting Codex login, including local Docker and remote servers.
- Submit through the authenticated Web origin to the current account's own operation; preserve state validation, PKCE, original redirect URI and owner-bound credentials.
- Parse only expected loopback callback URLs; never fetch user-supplied URLs, echo credentials, persist the pasted address or include it in a request query.
- Show redacted actionable errors, clear the input after submission and on operation changes, and retain existing automatic/tunnel flows.
- Cover malformed URLs, duplicate OAuth parameters, wrong-account state, expired operations, learner access and partner refusal.

No real authorization code or OAuth state is included in this report.

### Implementation

PR #1253 implements the recovery flow and regression coverage. Deployed to the local 3782 instance; container healthy, backend ready, unauthenticated completion requests rejected with HTTP 401. Existing reading fixes and user data are retained. A fresh user authorization is still needed to verify the live account connection.

evan188199-tech:
### Problem and behavior

A browser accessing a local Docker deployment cannot reach the container's localhost OAuth callback listener. The standard manifests do not reserve 1455/1457, and the settings card hides remote guidance on localhost, leaving sign-in waiting after browser authorization.

Add a callback completion form whenever Codex sign-in is waiting. Users paste the failed callback page's complete address into the same account's settings and choose Complete sign-in; no port edits, published callback ports or container restart are needed. Existing native callbacks, remote SSH guidance and the optional temporary Compose bridge remain available.

The authenticated POST endpoint parses expected loopback callback URLs without fetching them, sends them only to the caller's owner-scoped service, and retains state validation, original redirect URI and PKCE exchange. Malformed/duplicate parameters and cross-account state are rejected with redacted errors. The password-style field is cleared after submission and on operation changes; neither browser storage nor URL query transport is used. Partner identities remain denied.

Include the existing closed-dialog lazy-loading prerequisite in a separate commit so the dev baseline's knowledge-base/co-writer performance failures do not block this fix. No budgets are increased.

### Validation

- 154 OAuth service/API tests passed, including loopback listeners, callback URL validation, cross-account rejection, and existing token-exchange tests.
- Two rendered recovery tests passed on localhost: completion and rejected-state feedback, with field clearing.
- Web Node tests, unit suite, contracts, architecture, typecheck, lint and i18n checks passed; production build and all performance budgets passed.
- Real account OAuth authorization has not been replayed. A fresh user login is required to verify the final account connection after deployment.

Fixes #1252. Complements the temporary bridge from #743 and per-user scope from #781.

### Local deployment

Deployed as `deeptutor-local:codex-callback-recovery` over the existing reading-fixes image, retaining data and the previous image/config for rollback. Container health and backend readiness pass; the completion endpoint rejects unauthenticated requests with HTTP 401. Full GitHub Python/Web checks are still running; import, lint and generated-file checks have passed.

### DIFF
--- a/CONTAINERIZATION.md
+++ b/CONTAINERIZATION.md
@@ -90,6 +90,14 @@ Notes:
   the name. The `deeptutor-data` volume keeps your settings and workspace
   across restarts.
 
+### Complete Codex sign-in without a callback port bridge
+
+If the browser cannot open `http://localhost:1455/auth/callback` (or port `1457`), return to the same DeepTutor account's Codex settings while the login is still waiting. Copy the complete address from the failed callback page into **Callback URL**, then choose **Complete sign-in**. This works for local containers and remote deployments without publishing callback ports or restarting the service.
+
+The address is submitted to the authenticated account's existing login operation. The server parses it without fetching it, validates OAuth state, and retains the original redirect URI and PKCE exchange. The input is cleared after submission and is not saved in browser storage. Do not share the callback address in issues or logs. If the login expired, start again; a callback from another account's login is rejected.
+
+The temporary bridge below remains available when automatic browser return is preferred.
+
 ### Temporary local Codex OAuth bridge
 
 OpenAI Codex redirects the browser to fixed loopback ports `1455` or `1457`.
--- a/deeptutor/api/routers/settings.py
+++ b/deeptutor/api/routers/settings.py
@@ -770,6 +770,24 @@ async def start_openai_codex_oauth() -> dict[str, Any]:
         raise _codex_http_exception(exc) from None
 
 
+@router.post("/providers/openai-codex/oauth/complete")
+async def complete_openai_codex_oauth(request: Request) -> dict[str, bool]:
+    _require_codex_oauth_actor()
+    try:
+        payload = await request.json()
+    except ValueError:
+        payload = None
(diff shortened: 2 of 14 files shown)


evan188199-tech:
Superseded by the direct fix already landed on `dev` in commit `01615d6c97494d49fb07b50806b26fae416c1602` (referenced in issue #1252).

The core Docker OAuth callback recovery is now implemented on `dev`, so this PR should not be merged as a duplicate. It also contains additional tests/docs/performance changes; please cherry-pick or re-submit any of those separately if they are still needed.

PR #1279, opened 2026-09-07
fix: require recorded learner evidence before awarding mastery

1vivy:
### Description
Mastery tools can currently grade a model-supplied answer before a learner has replied, or pass a concept/design objective without any recorded assessment. This lets a tutor claim mastery based on prior chat or its own examples.

Require an answered interaction with a nonempty persisted learner reply before grading. Qualitative assessment additionally requires the same objective and rubric feedback, retains the question/learner answer/feedback as evidence, and resolves the interaction atomically. Deterministic grading leaves conceptual answers for rubric assessment. Update both language prompts and tool descriptions to pose explanation cards before assessing. Unambiguous choices typed in the composer are committed by the request runtime; a model's tool argument cannot replace an unreadable stored reply. Explicit learner correction still works.

Live validation also exposed repeated `mastery_status` calls consuming a whole context window. The prompt now requests status once per learner turn and reuses it until an action changes progress; pending-state instructions no longer suggest grading from earlier conversation.

An additional reproduction found that Settings saved a new default model while an existing worker kept using its cached old model. Turn activation now resolves the current catalog even without an explicit selection; the selected configuration stays scoped for the turn. Regression tests cover a warm worker, subsequent default changes, and explicit session selection with scope restoration.

Also fix two related failures found while reproducing this flow: accept the quiz tool's normalized `short_answer` type and use exact text or finite decimal equality instead of fuzzy matches (`0.05` must not equal `0.5`; `dependent` must not equal `independent`). Treat an index/retrieval error as unavailable material instead of using its error message as route grounding.

New outline conversations also sent their opening before the route-owned draft committed, using generic chat or a previous topic’s materials. Wait for the configured draft and matching topic, then carry the selected mastery session mode through the typed turn request. Two rendered integration regressions exercise the real provider, route hook and wire builder, including Strict Mode and an asynchronous topic switch; both fail on the previous code and pass with the fixes. Typecheck and targeted ESLint pass.

Paused-intake recovery: hydrate the saved mastery mode and re-subscribe to `waiting_input` turns, preserving the unanswered card across reopen. Checkpoint events before waiting, and retain Redis event payloads/sequence keys for the lifetime of the owner lease. Cancel and await the capability child when the consuming stream closes so a stopped question does not leave a background coroutine. New regressions reproduce missing hydration/replay, undurable paused questions and expiring sequence keys on the prior code. Six rendered tests, TypeScript, targeted ESLint, 47 runtime/subscription/real-Redis tests and 16 orchestrator tests pass. Full pre-commit still reports only the two baseline MyPy errors.

### Related Issues
No existing issue linked. Reproduced in a deployed course-learning workflow and verified against current `dev` (42fab3cf).

### Module(s) Affected
- [x] `services`
- [x] `tools`
- [x] `tests`
- [x] Other: `learning`, mastery prompt packs

### Checklist
- [x] I have read and followed the contribution guidelines; this PR targets `dev`.
- [x] I have updated the documentation (if necessary): tool contracts and both language prompts.

### Additional Notes
Validation: **538 passed, 1 deselected** for `deeptutor/learning/tests` plus `tests/services/session/test_turn_runtime_subscribe.py`, excluding `test_reconnect_after_turn_completion_still_carries_message_ids`. The excluded workspace-configuration failure also occurs on unchanged `dev`. Existing grading fixtures now simulate a real runtime answer before invoking the grading tool; new negative tests ensure model-only, empty, wrong-objective, and unreadable replies earn no credit.

`ruff check .` passes. `pre-commit run --all-files` was run on both this branch and unchanged `dev`: every hook passes except the same two pre-existing MyPy errors in `services/workspace/service.py:686` and `services/config/readiness.py:786`. `ruff check .` and `ruff format --check .` now both pass; two whitespace-only fixes to embedded Python examples in the built-in docx/xlsx skills are included because the repository-wide format gate checks them. Broader mastery capability tests have seven pre-existing failures without configured model/workspace state, reproduced on unchanged `dev`; no external model calls were used for these checks.

No course content, deployment configuration, credentials, or learner history is included. This prevents evidence-free credit; it does not replace the tutor's qualitative rubric judgement with an independent examiner.

CI follow-up: lint and all six import checks pass. The web job built successfully but fails existing route-size budgets (`/` 305/300 KB, `/co-writer` 324/320 KB, document editor 525/515 KB). The initial patch changed no web source, lockfile, or build configuration; the subsequent answer-card fix changes one expression in MasteryQuestionCard and adds rendered tests. Python 3.11, 3.13, and 3.14 passed the full suite. The 3.12 job passed 6762 tests and failed three prompt fixtures whose paths depended on the process working directory. Those fixtures now resolve paths from __file__; all 20 mastery-mode tests pass from an unrelated temporary working directory. The updated matrix is running.

Default-selection follow-up: 72 model-selection, WebSocket runtime, and partner runtime tests pass with an explicit temporary workspace root. Full pre-commit was rerun and still reports only the two baseline MyPy errors above.

Isolated live-provider validation on the patched 1.6.5 image: after switching the saved default from a gateway alias to direct Sonnet, fresh procedure and concept question cards appeared. Synthetic wrong answers were persisted, rejected by deterministic grading / qualitative assessment respectively, and followed by a new card. No actual learner progress was used. Immediate replies also reproduced a session-lock race: DONE previously preceded title generation, while the session/path locks were released only afterward. Terminal turns now cancel renewal and release input locks before publishing DONE, retaining the fencing token for post-turn metadata. A coordinated runtime regression verifies a successor can acquire the session at DONE and survives predecessor cleanup. The expanded runtime/model/partner suite passes 102 tests.

Final candidate verification repeats both isolated real-provider cycles with an immediate reply on DONE (no title-settlement wait): both pass through corrective grading/assessment and a fresh follow-up card. The final pre-commit run still reports only the two previously documented baseline MyPy errors.

Browser validation exposed a separate short-answer bug: the textarea was visible when there were no choices, but `answer` still used the empty picked choice until `freeSelected` was set. The card now uses the textarea whenever no choices exist. Three rendered regressions cover free text, choice labels, and explicit free-text alternatives (six tests pass including the neighboring card suite); typecheck and the production frontend build pass. Qualitative assessments now emit the common mastery_grade card verdict, retaining question ID, actual learner text and rubric feedback across reloads (73 mastery-tool tests pass).

Live browser verification after deployment: a previously disabled free-text Submit button enabled, delivered the learner-authored draft unchanged, displayed rubric feedback on the original card, and presented a new question. The next question was left for the learner.

Paused-intake deployment verification: the final question card (nested tool_metadata.ask_user) is durable before learner input; all 135 published events are persisted during the pause. Full browser reload restores the same card and Outline mode. Redis owner renewal retains the event stream, payload map and sequence counter. The real cancellation path now finishes and persists its trace. Final runtime/orchestrator/subscription tests: 62 passed; real Redis: 3 passed; rendered frontend: 6 passed.


r-rhythm:
### Additional repro for this fix: empty-args `mastery_assess` fires `mastery.assessed` on a path re-bound after chat deletion (v1.6.5)

This PR is exactly the right shape. Adding a concrete production repro that strengthens the case for requiring evidence at the engine boundary (not just in prompts/tool descriptions):

**Sequence**

1. A mastery path was carried by chat session A (09-07 evening → 09-08 morning; three KPs were marked via evidence-free `mastery_assess` calls in that session — same exploit this PR closes).
2. User deleted session A's chat in the morning (path state survives, it lives in the mastery store; only the transcript/context is gone).
3. A new session B re-attached to the path in the afternoon. Within its first graded turn, the model called `mastery_assess` with `args: {}` — no interaction id, no recorded learner reply — and the engine emitted `mastery.assessed {passed: true}` for another KP (revision bump visible in `mastery_events`).

**Why chat deletion made it worse**

After the cold re-bind, the tutor's transcript no longer contained the history behind the path's state (which KPs were "mastered" and why). The model then spent multiple turns improvising explanations for a state it couldn't reconstruct — hallucinated content, re-presented already-graded questions, self-contradictory turns — and each improvised `mastery_assess`/`mastery_grade` call further diverged the state machine from anything a learner could recognize. So the evidence-free path isn't just "premature mastery": with no transcript continuity it actively compounds state corruption.

**Two notes for the implementation**

1. The empty-args call is another instance of the long-context empty-tool-args failure mode reported in #779 (reproduced on deepseek models under heavy tool surfaces). Validation at the tool boundary catches it cleanly — one more reason this fix belongs in the engine, not the prompt.
2. Please also cover the `learner_mastery_overrides` endpoint in the same audit trail thinking: the learner-facing override API (`POST /topics/{path_id}/objectives/{kp_id}/override`) is a legitimate evidence-free claim channel by design (we used it to repair the wrongly-docked KP), and it emits `mastery.overridden` — good. Just make sure #1279's "recorded learner evidence" requirement stays scoped to the assess/grade tools so the override channel isn't accidentally locked out.

Verified: v1.6.6 does not change the mastery engine's grading/assess code (diffed the wheels), so the empty-args path should still reproduce there.

Issue #1245, opened 2026-09-05
Bug: question-notebook upsert with base64 images fails with OperationalError: 22 values for 23 columns


larryluozhang:
## Environment
- DeepTutor v1.6.4 (ghcr.io/hkuds/deeptutor:latest)

## Reproduction
POST /api/question-notebook/entries/upsert with a non-null `user_answer_images` field (base64 image) returns HTTP 500:

```
{"detail":"OperationalError: 22 values for 23 columns","type":"OperationalError"}
```

The same payload with `user_answer_images: null` succeeds (entry created). Looks like the INSERT into the images/attachments table passes one fewer value than the column count — likely a missing column in the SQL statement (e.g. a `created_at`/owner/scope column not included in the values list).

## Workaround
Retry the upsert without images (entries save fine, images lost).

Reproducible against any session; happened with both JPEG and PNG base64 payloads.

wangyuxy9805-web:
Hi, I'd like to work on this. I reproduced the failure on current `dev` (`269f65d`). The image-specific `INSERT` lists 23 columns but supplies only 22 SQL values: it is missing the `resolved` placeholder before the literal `bookmarked`/`followup_session_id` values, although the bound tuple already provides the resolved value.

I'll submit a focused fix plus an API regression test that upserts a base64 image and verifies the persisted image metadata. Disclosure: I'm using AI assistance, and I will manually review and run the tests before submitting.

pancacake:
Thanks for your comments！Will be fixed in the next version～

Fixed on `dev` in #1258. The root cause was one missing placeholder — the images branch listed 23 columns against 22 values — so both INSERT branches are now 23/23. `ai_judgment` was not the missing piece; it already has a `DEFAULT` in the schema.

PR #1283, opened 2026-09-07
fix(session): include ai_judgment in notebook_entries INSERT (#1245)

sjagadee:
## Problem

[Bug #1245](https://github.com/HKUDS/DeepTutor/issues/1245): `POST /api/question-notebook/entries/upsert` with `user_answer_images` returns HTTP 500 `OperationalError: 22 values for 23 columns`.

PR #1258 added the missing `?` placeholder for `bookmarked` / `followup_session_id` in `_upsert_notebook_entries_sync`, but the table schema has a 24th column, `ai_judgment`, added by `_migrate_notebook_entries_add_ai_judgment` (which runs unconditionally on every connection). Both INSERT branches still omit `ai_judgment`, so any database that has run the migration — including a fresh pull of `ghcr.io/hkuds/deeptutor:latest` — fails the same way.

## Fix

Bring the INSERT column lists and value tuples into sync with the post-migration schema in both branches of `_upsert_notebook_entries_sync`:

- Add `ai_judgment` to each column list.
- Bind `""` in each value tuple (consistent with the inline `0` / `''` already used for `bookmarked` and `followup_session_id`).
- Add the one missing `?` in the with-images VALUES clause that the column-add exposed.

## Verification

- New test `test_upsert_notebook_entries_with_answer_images` exercises both INSERT branches against the real `SQLiteSessionStore`.
  - Reverting just the production change and re-running the test reproduces the exact reported error: `sqlite3.OperationalError: 22 values for 23 columns`.
  - With the fix applied, the test passes and verifies that:
    - A fresh upsert carrying `user_answer_images` persists the image list.
    - A re-upsert without images preserves the previously-stored image list (the no-images branch must not clobber `user_answer_images_json`).
- Full `test_sqlite_store.py` (32 tests) and `test_notebook_router.py` (15 tests) pass.

## Files

- `deeptutor/services/session/sqlite_store.py` (+6 / −4)
- `tests/services/session/test_sqlite_store.py` (+63)

## Notes

- This is a follow-up to #1258; that PR is still mergeable and ships complementary work (image attachment store plumbing) but is not strictly required to land this fix.
- No new dependencies; no schema changes; no migration changes.

### DIFF
--- a/deeptutor/learning/tests/test_mastery_mode.py
+++ b/deeptutor/learning/tests/test_mastery_mode.py
@@ -395,11 +395,26 @@ async def test_a_built_goal_reports_its_identity_too(path_id):
 
 
 def _pack(language: str) -> dict:
+    from pathlib import Path
+
     import yaml
 
-    return yaml.safe_load(
-        open(f"deeptutor/capabilities/mastery/prompts/{language}/mastery_loop.yaml")
-    )
+    # The fixtures are checked in next to the source tree, not next to the
+    # test directory — and pytest invokes this from whatever CWD was set when
+    # the run started (the repo root in CI, ``cwd`` from a developer shell
+    # otherwise). Anchor on this file's location so both work.
+    repo_root = Path(__file__).resolve().parents[3]
+    with open(
+        repo_root
+        / "deeptutor"
+        / "capabilities"
+        / "mastery"
+        / "prompts"
+        / language
+        / "mastery_loop.yaml",
+        encoding="utf-8",
+    ) as handle:
+        return yaml.safe_load(handle)
 
 
 def test_study_never_offers_to_do_the_reviewing_itself():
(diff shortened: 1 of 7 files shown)


pancacake:
Thanks for digging into #1245 — your diagnosis of the crash was right, and your regression test was the best part of this PR.

We landed the smaller fix from #1258 instead: the images branch was missing a single placeholder (23 columns, 22 values), so adding one `?` brings both INSERT branches back to 23/23 without restating the schema's `DEFAULT ''` in Python. Your store-layer test has been carried over to `dev` in e5a7ff47 with you as co-author — it now guards both branches, including the re-upsert that must not clobber stored images.

The mastery-prompt cwd fix and the SKILL.md formatting fixes in here landed via #1282, and the route-budget increases turned out to be unnecessary: #1274 pulled the ~95KB generated brand-icon table out of the shared app-shell chunk, so `/` is now 202KB against the original 300KB budget.

Closing as superseded. Thanks again!

PR #1311, opened 2026-09-08
feat(codex-auth): discover stable client version (#1296)

### LINKED ISSUE #1296: [Feature Request]: Automatically discover the Codex client version for model catalog refresh

- [x] I have searched the existing feature requests and pull requests and found no direct duplicate of this feature as of September 8, 2026.
- [x] I believe this is a legitimate feature request, not just a question or bug report.

### Feature Request Description

Automatically discover a stable Codex client version when signing in to OpenAI Codex or manually refreshing its model catalog, rather than relying exclusively on a hard-coded compatibility version.

PR #1264 addressed missing eligible GPT-6 Astra entries by updating the fixed catalog request version to `0.153.4`. This request is a follow-up enhancement: reduce the need for future manual version bumps when the account catalog changes with the advertised client version.

DeepTutor should read the latest stable `@openai/codex` package version from the official npm registry and use it as the catalog request's `client_version`.

This should be metadata-only. It must not require, install, download, or upgrade the Codex CLI, or read its local login state.

### Related Module

API/Backend — OpenAI Codex OAuth and model catalog services.

### Use Case

A user signs in with their own Codex account and periodically uses **Refresh models** to discover available models.

A fixed catalog client version can omit models that are eligible for that account. Updating DeepTutor's compatibility constant addresses the immediate case, but requires another code change whenever a similar version-dependent catalog change occurs.

Automatic version discovery would let an explicit refresh follow the upstream stable client version while preserving account isolation, predictable failure handling, and the user's existing model selection.

### Additional Context

Related work: #1264 provides the fixed-version compatibility update. This proposal adds automatic discovery and fallback behavior on top of that fix.

#### Proposed behavior

- Discover the stable client version only during sign-in and explicit model refresh.
- Isolate npm requests from OAuth credentials, account headers, cookies, and client authentication.
- Bound discovery with a three-second deadline, a 64 KiB response limit, and no redirects.
- If discovery fails, use the account's last successfully used version, then the built-in fallback if no history exists.
- Persist a newly discovered version only after a valid catalog response.
- Retry the previous version at most once, and only for explicit version rejection or an incompatible catalog structure.
- Keep authentication, rate-limit, TLS, and malformed-response failures distinct from version incompatibility.
- Never reuse catalog validators across client versions or credential generations.
- Preserve successful version history across same-account credential renewal; clear it on logout.
- Prevent late responses from restoring a logged-out cache or overwriting a newer login's cache.

#### Acceptance criteria

- [ ] Sign-in and manual refresh can use a dynamically discovered stable client version.
- [ ] Discovery failure follows the documented fallback order.
- [ ] npm requests contain no OAuth credentials and obey the response/time limits.
- [ ] Cache and version history remain isolated across accounts, credential changes, and logout.
- [ ] Newly discovered models do not replace an existing user-selected model.
- [ ] Status reads and inference do not query npm.
- [ ] Tests cover version discovery, failure classification, fallback, and concurrent cache publication.

#### Non-goals

- Installing or updating the Codex CLI.
- Background polling or a user-facing version setting.
- Granting access to models unavailable to the signed-in account.
- Guaranteeing inference compatibility merely because a model appears in the catalog.

evan188199-tech:
Closes #1296

## Summary
- Add metadata-only discovery of the stable `@openai/codex` npm version for sign-in and explicit model refresh.
- Use a private HTTP client for npm requests with no OAuth credentials, cookies, or account headers.
- Bound discovery to 3 seconds, 64 KiB, stable semantic versions, and no redirects.
- Persist a discovered version only after a valid live catalog response.
- Add account-bound version history with fallback to the last successful account version, then the built-in `0.153.4`.
- Retry the previous version at most once for explicit version rejection or incompatible catalog structure.
- Keep authentication, rate-limit, availability, malformed-response, and version-incompatibility failures distinct.
- Guard catalog caches, version history, invalidation, and managed-profile publication by credential generation and account binding.
- Clear version history and catalog cache on logout while preserving history across same-account credential renewal.
- Preserve the existing model selection when refreshing managed Codex models.

## Verification
- [x] `pytest tests/services/codex_auth --ignore=tests/services/codex_auth/test_oauth.py` - 140 passed
- [x] `pytest tests/services/codex_auth/test_oauth.py` - 15 passed
- [x] `ruff check` - passed
- [x] `ruff format --check` - passed
- [x] `git diff --check` - passed
- [x] Live npm metadata smoke test: HTTP 200, 3,521 bytes, 0 redirects, `@openai/codex` `0.153.4`

## Security notes
- npm discovery does not read or install the local Codex CLI.
- npm requests are metadata-only and isolated from OAuth client state.
- Account history stores only hashed account bindings and version strings.
- Status reads and inference do not query npm.

### DIFF
--- a/deeptutor/services/codex_auth/catalog.py
+++ b/deeptutor/services/codex_auth/catalog.py
@@ -20,13 +20,18 @@ from .constants import (
 )
 from .contracts import CatalogSnapshot, CodexAuthError, CodexCredentials, CodexModel
 from .storage import CodexCredentialStore
+from .version import validate_stable_version
 
 
-def parse_models_response(payload: Mapping[str, Any]) -> tuple[CodexModel, ...]:
+def parse_models_response(
+    payload: Mapping[str, Any],
+    *,
+    incompatible_code: str = "catalog_invalid",
+) -> tuple[CodexModel, ...]:
     raw_models = payload.get("models")
     if not isinstance(raw_models, list):
         raise CodexAuthError(
-            "catalog_invalid",
+            incompatible_code,
             "Codex returned an invalid model catalog.",
             502,
         )
@@ -41,7 +46,7 @@ def parse_models_response(payload: Mapping[str, Any]) -> tuple[CodexModel, ...]:
     for raw_model in raw_models:
         if not isinstance(raw_model, dict):
             raise CodexAuthError(
-                "catalog_invalid",
+                incompatible_code,
                 "Codex returned an invalid model catalog.",
                 502,
             )
@@ -51,7 +56,7 @@ def parse_models_response(payload: Mapping[str, Any]) -> tuple[CodexModel, ...]:
         display_name = raw_model.get("display_name")
         if not isinstance(slug, str) or not slug:
             raise CodexAuthError(
-                "catalog_invalid",
+                incompatible_code,
                 "Codex returned an invalid model catalog.",
                 502,
             )
@@ -132,7 +137,9 @@ class CodexModelCatalog:
         self,
         credentials: CodexCredentials,
         force: bool,
+        client_version: str | None = None,
     ) -> CatalogSnapshot:
+        requested_version = validate_stable_version(client_version or CODEX_CLIENT_VERSION)
         now = int(self._clock())
         account_hash = hashlib.sha256(credentials.account_id.encode("utf-8")).hexdigest()
         cache = self._matching_cache(credentials, account_hash)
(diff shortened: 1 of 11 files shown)


ZJPex:
Hi, I noticed that this PR also addresses #1296. I opened that issue and commented on September 8 that I wanted to take it on. My implementation was already in progress, but publishing it was delayed by a GitHub authorization issue. It is now available in #1317.
Given my earlier comment expressing my intention to work on this, could a maintainer please review my PR #1317 first, before deciding which implementation to proceed with? I've also added a coordination note to #1296. Thank you.

evan188199-tech:
Thanks @ZJPex, and sorry for the overlap. Your earlier interest and #1317 make sense as the main path for this change. I'll close #1311 to keep the work on a single clear track. If any tests or implementation details from #1311 are useful later, I'm happy to contribute them to #1317 as a focused follow-up with proper credit. Thanks for taking this on.

ZJPex:
Thanks for understanding! I really appreciate you closing #1311 to avoid duplicate work, and also your offer to contribute any useful tests or implementation details to #1317.

I'll take a closer look at your implementation as well, and if there are parts that could strengthen #1317, I'd be happy to coordinate and incorporate them with proper credit.

Thanks again!

PR #1313, opened 2026-09-09
fix(chat): always clear sending state in useCardSubmission (#1289)

### LINKED ISSUE #1289: [Bug]: v1.6.5 Disabled submit button in Mastery Path mode

### Describe the bug

In Mastery Path mode, after typing an answer inside the question input box, the submit button stays greyed‑out and cannot be clicked. Answers can only be submitted via the bottom chat input box.

### Steps to reproduce

Version: v1.6.5

### Expected Behavior

_No response_

### Related Module

Dashboard

### Configuration Used

_No response_

### Logs and screenshots

_No response_

### Additional Information

- DeepTutor Version:
- Operating System:
- Python Version:
- Node.js Version:
- Browser (if applicable):
- Related Issues:

sjagadee:
## Summary

Fixes #1289 — Submit button stays permanently disabled in Mastery Path mode after typing an answer in the question input box.

## Root Cause

In `web/hooks/use-card-submission.ts`, the `submit` callback only called `setSending(false)` when `onSubmit` returned **explicitly `false`**. When `answerMasteryQuestion` calls `sendMessage(...)` (fire-and-forget) and returns `true`, the `accepted === false` branch was never entered — `sending` stayed `true` forever, locking all card controls including the Submit button (`disabled={locked || !answer}` where `locked = settled || sending || skipping`).

## Fix

Move `setSending(false)` **outside** the `if (accepted === false)` guard so it always clears the sending state after `onSubmit` resolves. `setFailed(true)` remains conditional — only an explicit `false` from the host means the submission was rejected and the card must reopen.

```diff
- if (accepted === false) {
-   setSending(false);
-   setFailed(true);
- }
+ setSending(false);
+ if (accepted === false) {
+   setFailed(true);
+ }
```

## Testing

- Browser-tested: Submit button now unlocks immediately after dispatching a mastery question answer
- Existing `ask-user-terminal-turn.spec.tsx` confirms empty-submit is still rejected while awaiting reply
- Hook behavior unchanged for the `accepted === false` path (failed state set, card reopens)

### DIFF
--- a/web/hooks/use-card-submission.ts
+++ b/web/hooks/use-card-submission.ts
@@ -38,8 +38,10 @@ export function useCardSubmission(onSubmit: SubmitUserReply) {
       // ``undefined`` is a host that does not report a verdict; only an
       // explicit ``false`` reopens the card. The picked answers survive in the
       // caller's state, so retrying is one click.
+      // Clear ``sending`` unconditionally: truthy or undefined both mean the
+      // host accepted the payload and the loading state must end immediately.
+      setSending(false);
       if (accepted === false) {
-        setSending(false);
         setFailed(true);
       }
     },


pancacake:
Two problems. First, it does not fix #1289: the free-text submit bug is that `answer` reads `freeSelected ? freeText.trim() : picked` while the textarea renders on `freeSelected || !hasChoices` — the real fix is #1291, now merged. Second, clearing `sending` unconditionally opens a double-submit window: `locked = settled || sending || skipping` and `settled = Boolean(grade)`, but a mastery grade only arrives a whole turn later, so Submit becomes clickable again and the same answer can start a second turn (`AskUserOptions` has the same shape via `locked = submitted || streaming`). It also contradicts the documented contract in `useCardSubmission` that only an explicit `false` reopens a card. The spinner-that-never-stops you were aiming at is real and we are tracking it — the fix needs to distinguish `false`/`undefined` (reopen) from `true` (stay locked until `answered`).

Issue #1316, opened 2026-09-09
[Bug]:Book spine synthesis fails with "LLM returned empty response" when using a reasoning model via an Anthropic-format endpoint


zhiyuan8015:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

When generating a book, the **spine synthesis** stage (`SpineSynthesizer`) logs:

生成书籍时，**spine 合成**阶段（`SpineSynthesizer`）打出如下日志：

```
WARNING deeptutor.Book.spine_synthesizer - LLM returned empty response
```

The book pipeline then silently degrades to a single placeholder "Overview" chapter because the LLM returned no usable content.

随后书籍流水线因为拿不到可用的 LLM 内容，静默退化为单个占位章节「Overview」。

**Key point / 关键点**: the log does **not** contain `SpineSynthesizer LLM call ... failed` or `LLM call failed`, which means the HTTP request **succeeded** but the model returned an **empty text body** (not an exception).

**关键点**：日志里**没有** `SpineSynthesizer LLM call ... failed` 或 `LLM call failed`，说明 HTTP 请求是**成功**的，但模型返回了**空的文本正文**（不是抛异常）。

### Steps to reproduce

1. Configure an LLM profile with `binding=custom_anthropic`, `base_url=https://api.deepseek.com/anthropic`, and a reasoning model such as `deepseek-v4-pro`. / 配置 `binding=custom_anthropic`、`base_url=https://api.deepseek.com/anthropic`，模型使用推理模型 `deepseek-v4-pro`。
2. Create/confirm a book so the engine runs `confirm_proposal → SpineSynthesizer.synthesize`. / 创建/确认一本书，触发 `confirm_proposal → SpineSynthesizer.synthesize`。
3. Observe the warning and the resulting single-chapter spine. / 观察该警告以及退化为单章节的目录。
4. Contrast: sending a short chat message to the same model returns a normal reply. / 对照：用同一模型发一条简短的聊天消息，能正常收到回复。

### Expected Behavior

The spine stage should produce a valid JSON payload `{concept_graph, chapters}` and a full chapter spine.

spine 阶段应产出合法的 JSON 载荷 `{concept_graph, chapters}`，以及完整的章节目录。

### Related Module

Book

### Configuration Used

[...]

PR #1320, opened 2026-09-09
fix(config): let a book capability's agents.yaml settings take effect

AmirF194:
`get_agent_params("book")` never reads `agents.yaml`. `section_map` in `deeptutor/services/config/loader.py` maps module names to their `agents.yaml` section, and "book" was never added to it (the four book agents in `deeptutor/book/agents/` all call `get_agent_params(module_name="book")`). Without an entry, the function returns the bare global defaults (`temperature=0.5`, `max_tokens=4096`) before it ever opens the file, so a `capabilities.book.max_tokens` set by the user has no effect. That lines up with the "LLM returned empty response" failures with a reasoning model: there's no way to raise the token budget for book synthesis at all.

Added `"book": ("capabilities", "book")` to `section_map`, and a matching `"book"` entry to `DEFAULT_AGENTS_SETTINGS["capabilities"]` in `deeptutor/services/setup/init.py` so it seeds the same way `solve`/`research`/etc. do. Kept the seeded default at 0.5/4096, same as what the code silently returned before, so this only changes behavior for someone who actually sets `capabilities.book` in their `agents.yaml`.

Scoped to this one config-plumbing gap. The issue also describes two other causes (a provider-format extraction issue and a reasoning-model token-budget-sharing issue); didn't touch either, those need their own fix and tests.

Added `tests/services/config/test_book_capability_config.py`: one test pins that a configured `max_tokens`/`temperature` is honored (fails on `dev`, passes here), one pins that the no-override default is unchanged, one asserts `DEFAULT_AGENTS_SETTINGS` actually carries the `book` entry. Ran the full suite on Python 3.12 in Docker: 6775 passed, 47 skipped, same 8 failures as on `dev` (missing `git`/`libmagic` in the container, a sandbox cgroup test), none touching config or setup. `ruff check`, `ruff format --check`, `lint-imports`, and `scripts/check_architecture.py` all pass.

### DIFF
--- a/deeptutor/services/config/loader.py
+++ b/deeptutor/services/config/loader.py
@@ -236,6 +236,7 @@ def get_agent_params(module_name: str) -> dict:
         "question": ("capabilities", "question"),
         "co_writer": ("capabilities", "co_writer"),
         "visualize": ("capabilities", "visualize"),
+        "book": ("capabilities", "book"),
         "brainstorm": ("tools", "brainstorm"),
         "vision_solver": ("plugins", "vision_solver"),
         "math_animator": ("plugins", "math_animator"),
--- a/deeptutor/services/setup/init.py
+++ b/deeptutor/services/setup/init.py
@@ -75,6 +75,7 @@ DEFAULT_AGENTS_SETTINGS = {
         "question": {"temperature": 0.7, "max_tokens": 4096},
         "co_writer": {"temperature": 0.7, "max_tokens": 4096},
         "visualize": {"temperature": 0.4, "max_tokens": 16384},
+        "book": {"temperature": 0.5, "max_tokens": 4096},
         "chat": {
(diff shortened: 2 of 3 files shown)


pancacake:
The wiring here is right and it is landing in v1.6.7, but not as this patch, because adding `book` to `section_map` alone does not fix #1316. Once the map entry exists, `get_agent_params("book")` still finds no `book` key in `DEFAULT_AGENTS_SETTINGS` and falls back to the same 4096 `global_defaults` — nothing changes for a user who never edited agents.yaml. `dev` now adds the map entry *and* seeds `book` at 12000, and pairs it with a low-effort retry for the case a reasoning model spends the whole budget on hidden tokens, which is what actually produced the one-chapter book. Thanks for finding the missing map entry.

PR #1225, opened 2026-09-04
fix(learner): honor web learning policy (#1222)

evan188199-tech:
## Summary

- Preserve `learning_policy` in auth status and use `allowed_surfaces` to render only the learner-accessible sidebar destinations.
- Redact deployment-owned settings for learning accounts while retaining appearance, learner profile, and about; fall back to the narrow `/api/settings/ui` projection when `/api/settings` is denied.
- Keep chat history when optional course, topic, or reading-label requests fail, and handle denied memory and Co-Writer responses without crashing or burying the server detail.
- Expose the effective first granted-and-available model as the learner default and map exactly `/api/settings/llm-options` to the `chat` surface without opening the rest of the settings router.

Fixes #1222.

## Verification

- `DEEPTUTOR_TEST_REDIS_URL=redis://127.0.0.1:6379/0 ... -m pytest -q tests/multi_user/test_account_presets.py tests/multi_user/test_capability_access.py tests/multi_user/test_owner_bound_models.py tests/multi_user/test_personal_codex_models.py` — 30 passed.
- `ruff check` and `ruff format --check` on the changed Python files — passed.
- `npm run test:node` — 1041 passed.
- `npm run typecheck` — passed.
- `npm run lint` — 0 errors; 72 pre-existing warnings.
- `npm run i18n:check` — passed.
- `npm run build` — passed.
- `WEB_BASE_URL=http://127.0.0.1:4319 npx playwright test tests/e2e/learning-policy-shell.audit.ts --project=ui-audit` — 1 passed, covering desktop/mobile navigation, settings redaction, retained history, and assigned-model display.

The server remains the authority for learning-policy enforcement; the web shell only consumes the policy already returned by auth status. This PR intentionally excludes the separate Windows CI encoding issue and unrelated Playwright baseline failures.

Issue #1222, opened 2026-09-03
[Bug]: Learning accounts — the policy reaches the client but nothing reads it, and CI cannot see the fallout


khunmax2:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

Enabling a learning policy on an account (or creating a `learner` preset) leaves
the UI in a state where most of what it offers answers 403. The backend guard is
correct and consistent; the frontend was never taught that "denied" is a normal
state, so the restriction surfaces as errors after a click rather than as an
absence.

`GET /api/auth/status` already returns `learning_policy` with `allowed_surfaces`,
and `web/lib/auth.ts` types it. As of v1.6.4 the only reference to that field in
the whole frontend is its own type declaration — `useAuthStatus` drops it.

That is exactly the contract #1111 set out:

> "Expose the effective public policy through auth status **so clients can render
> the learner experience** without trusting client-side restrictions."

The policy is exposed. Nothing renders from it.

### What happens

Nine distinct symptoms. Grouped by cause rather than by screen:

**The policy is delivered and ignored**

1. The sidebar offers all twelve features. Co-Writer, Books, Mastery Path,
   Partners, My Agents, Learner Anima, Learning Space, Memory and Knowledge
   Center each answer 403 on click.
2. Settings renders Network, Models, Knowledge Base, Chat and Memory. Each
   fetches and fails — the visible result is `Failed to load tools: HTTP 403`
   and `Couldn't load capability settings`.
3. The Guardian section is shown whenever `preset` is `standard` or `custom`,
   without checking whether a learning policy is present. Everything it reads
   is under `/api/multi-user/*`, which is guarded.

**Code that assumes the request succeeded**

4. `web/features/settings/sections/MemorySettingsSection.tsx` does
   `.then((res) => res.json() as Promise<MemorySettingsDTO>)` with no `res.ok`
   check. On 403 the error body is cast to the DTO — the `as` hides this from
   `tsc` — the component's own `if (!settings)` guard passes because the object
   is truthy, and the first read of `settings.update.l2_budget` throws out of
   render. **The whole settings page unmounts to "This page couldn't load".**
   Reachable by scrolling to any anchor below Memory.
5. `SettingsStore` treats the 403 from `/api/settings` as fatal and paints
   `Settings fetch failed: HTTP 403` above controls that work — `/api/settings/ui`
[...]

PR #1323, opened 2026-09-09
fix(learner): complete surface map for first-party learner screens + flip preset on restrictions assignment (#1222)

larryluozhang:
## What this fixes

Two **server-side** gaps behind the learner-account 403 storm reported in #1222 (complements the frontend work in #1225 — the breakage is two-sided).

### 1. The surface map doesn't cover first-party learner screens

`_learning_surface_for_path` maps only five prefixes. The scoped learner shell (preset=`learner`) links **Knowledge Center** and **Mastery Path** in its nav, but their endpoints fall through to the default-deny and 403:

```
learner login → click 知识中心 (Knowledge Center)
  → GET /api/knowledge-bases → 403 "This learning account cannot use the requested server surface."
learner → click 精通之路 (Mastery Path)
  → GET /api/mastery-paths/topics/index → 403 (same)
```

Measured against a live deployment (v1.6.5 image) by driving the real learner UI and capturing every API call — full endpoint/status table in [this comment](https://github.com/HKUDS/DeepTutor/issues/1222#issuecomment-5538884494).

**Fix:**
- `/api/mastery-paths` → `chat`, **all methods**. Progress/objectives/sessions are the learner's *own* per-user data; the router already scopes every record to the current account, so `PATCH progress`, `POST redo`, etc. are legitimate learner activities and required for the feature to work at all.
- `/api/knowledge-bases` → `reading`, **read-only methods only** (`GET`/`HEAD`/`OPTIONS`). Browsing the assigned library is a legitimate learner activity; `POST`/`DELETE` still default-deny because they touch shared/admin-owned resources.

### 2. Admin-created learner accounts get a broken preset/policy mix

Accounts created via the admin Users API keep `preset: "standard"` when a guardian later assigns restrictions. The frontend keys the app shell off `preset`, loads the full admin UI, and *every* shell request 403s in a storm right after login (`/api/settings`, `/api/knowledge-bases`, `/api/mastery-paths/topics/index`, `/api/courses`, `/api/documents`, …).

**Fix:** `PUT /learners/{id}/restrictions` now treats "assigning restrictions" as the moment an account becomes a learning account — it seeds the default learning policy via `learner_grant()` (instead of 409ing) and flips the account preset to `learner`, so the frontend renders the scoped learner shell.

## Verification

- New regression tests: **22 passing** (`tests/multi_user/test_learning_surface_map.py`) — surface-map parametrization incl. mutation denial, plus the preset flip end-to-end.
- Existing `tests/multi_user/` suite: 197 passed / 12 failed; the 12 failures reproduce identically on unmodified `main` in a minimal venv (pre-existing, unrelated to this change).
- Live-deployment verification of the equivalent hot-patch: learner login, Knowledge Center, Mastery Path, immersive reading, chat — all functional with zero 403s; KB upload/delete still correctly 403 for the learner.

## Notes

- Frontend-side scoping (rendering from `learning_policy` instead of calling denied endpoints) is tracked in #1222 / PR #1225; this PR is the complementary server side.
- The default-deny philosophy is unchanged — this only admits the first-party surfaces the learner shell itself links to.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deeptutor/api/routers/auth.py
+++ b/deeptutor/api/routers/auth.py
@@ -401,7 +401,7 @@ async def require_admin(
     return payload
 
 
-def _learning_surface_for_path(path: str) -> str:
+def _learning_surface_for_path(path: str, method: str = "GET") -> str:
     normalized = "/" + str(path or "").lstrip("/")
     for root, surface in (
         ("/api/reading", "reading"),
@@ -410,9 +410,20 @@ def _learning_surface_for_path(path: str) -> str:
         ("/api/question", "chat"),
         ("/api/question-notebook", "chat"),
         ("/api/sessions", "chat"),
+        # Mastery Path progress/topics are the learner's own per-user data;
+        # the router already scopes every record to the current account, so
+        # all methods (including progress PATCH/POST) belong to "chat".
+        ("/api/mastery-paths", "chat"),
     ):
         if normalized == root or normalized.startswith(f"{root}/"):
             return surface
+    # Knowledge-center browsing is a legitimate learner activity, but KB
+    # mutations affect shared/admin-owned resources — read-only methods only.
+    if method.upper() in ("GET", "HEAD", "OPTIONS") and (
+        normalized == "/api/knowledge-bases"
+        or normalized.startswith("/api/knowledge-bases/")
+    ):
+        return "reading"
     return ""
 
 
@@ -424,7 +435,7 @@ async def require_learning_surface(
     from deeptutor.multi_user.learning_access import assert_learning_surface
 
     try:
-        assert_learning_surface(_learning_surface_for_path(request.url.path))
+        assert_learning_surface(_learning_surface_for_path(request.url.path, request.method))
     except PermissionError as exc:
         raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail=str(exc)) from exc
 
--- a/deeptutor/api/routers/multi_user.py
+++ b/deeptutor/api/routers/multi_user.py
@@ -45,6 +45,7 @@ from deeptutor.multi_user.identity import (
     list_user_info,
     set_book_permission,
     set_password,
+    set_preset,
 )
 from deeptutor.multi_user.knowledge_access import admin_kb_base_dir
 from deeptutor.multi_user.model_access import is_owner_bound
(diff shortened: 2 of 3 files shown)


larryluozhang:
Rebased onto current main (v1.6.9 layout — `_learning_surface_for_path` now lives in `deeptutor/api/routers/auth.py`; change content unchanged). Verified against main: all 22 regression tests in `tests/multi_user/test_learning_surface_map.py` pass, and the full `tests/multi_user/` suite shows no new failures (12 pre-existing failures on clean main in a minimal env, unrelated).

The gap is still present on main today: the map covers only reading/courses/chat/question/question-notebook/sessions, while the `_auth` guard now also wraps memory/dashboard/imports/capabilities routers — learner accounts still default-deny 403 on `/api/knowledge-bases/*` and `/api/mastery-paths/*`. This PR keeps the fix surgical: browse-only mapping for knowledge bases (GET/HEAD/OPTIONS), full mapping for mastery paths (per-user scoped data), and the preset flip that prevents the standard-preset+policy 403 storm.

larryluozhang:
Update after field-testing against v1.6.9 in production: the learner shell (preset=learner + policy) still fires read-only bootstrap probes at app-level prefixes (/api/settings, /api/tools, /api/partners, /api/partner-groups, /api/dashboard/suggestions, /api/capabilities/registered, /api/system/update, /api/subagents/settings). Each 403 surfaces the 'cannot use the requested server surface' banner. This revision maps GET/HEAD/OPTIONS on those prefixes to chat while keeping all mutations default-denied. 34 regression tests pass locally (22 original + 12 new).

larryluozhang:
One more from production: a learning account saving UI language/theme got 'Unsaved changes: HTTP 403'. interface.json resolves through the per-user path service into the caller's own workspace, so PUT/PATCH /api/settings/ui is safe to allow for learners. Added with 2 test cases (36 total passing).

larryluozhang:
Root-caused the language-save 403 in the browser (v1.6.9): 'Apply changes' on the General settings page fires a three-call chain — PUT /api/settings/draft → PUT /api/settings/ui → DELETE /api/settings/draft. The draft document is per-user, and catalog edits inside a draft are admin-gated by the endpoint itself, so the draft path is safe to open to learners. Added with tests; 40 cases passing. Verified live: learner account now saves interface + output language with 200s across the whole chain.

PR #1358, opened 2026-09-10
fix(rag): surface embedding connectivity failures (#1356)

### LINKED ISSUE #1356: [Bug]: Knowledge retrieval obscures embedding connectivity failures
### Describe the bug

When a selected knowledge base depends on an embedding service that is stopped or unreachable, the underlying RAG pipeline logs a clear connectivity error (for example, `Cannot connect to Ollama at ...`), but chat and Mastery Path turns can continue waiting without surfacing that cause as a clear retrieval result/error event.

Users then see only a generic long-idle timeout such as `connection timed out - no response received for 180 seconds`, which does not tell them to start the embedding service, check the endpoint, or re-index with a compatible embedding model.

### Steps to reproduce

1. Select a vector KB whose active embedding endpoint points to a local service such as Ollama.
2. Stop that embedding service.
3. Start a chat or Mastery Path turn that invokes KB retrieval.
4. The backend logs an embedding connectivity failure, while the client can remain in a quiet streaming state until its idle watchdog fires.

### Expected behavior

- KB retrieval performs a bounded embedding connectivity check before entering long-running retrieval/index work.
- Connection failures are returned as an actionable `rag` tool error containing the configured endpoint/provider and a short remediation hint.
- The turn emits a terminal failure/result event instead of leaving the UI waiting.
- Existing provider-specific error messages remain available in logs and diagnostics.

### Related module

Knowledge Base / RAG tool / embedding client

### Additional context

This is distinct from older KB upload progress bugs and older embedding model/version migration issues: the embedding endpoint is unavailable, and the immediate problem is chat-tool observability and turn termination.

evan188199-tech:
## Summary
- Normalize common embedding connectivity failures (`connection refused`, `cannot connect`, Ollama/embedding timeouts, etc.) to `error_type=embedding_connectivity`.
- Preserve the provider and endpoint details in the actionable RAG error message.
- Mark failed RAG tool results as unsuccessful, omit fabricated citation sources, and terminate the turn for embedding connectivity failures.
- Keep ordinary RAG errors non-terminal so the model can still handle them.

Closes #1356.

## Design note
This intentionally does not add a separate embedding preflight request before every retrieval. That would double embedding calls in the normal path. Instead, the first failing search response is normalized into an actionable terminal result. A dedicated health-check command can be layered on later if maintainers prefer explicit endpoint diagnostics.

## Validation
- PASS: `python -m pytest tests/services/rag/test_llamaindex_embedding_failures.py tests/core/test_builtin_tools.py` (30 passed, 2 warnings)
- PASS: `python -m ruff check` on all changed files
- PASS: `python -m ruff format --check` on all changed files
- PASS: `git diff --check`

A wider local run also included `tests/core/agentic/test_tool_dispatch_events.py` and `tests/agents/chat/test_kb_seed_context.py`. The latter already fails during test workspace initialization with `WorkspaceError: The selected workspace folder does not exist.`, independent of this change.

### DIFF
--- a/deeptutor/services/rag/pipelines/llamaindex/errors.py
+++ b/deeptutor/services/rag/pipelines/llamaindex/errors.py
@@ -10,6 +10,32 @@ def search_error_result(query: str, exc: Exception) -> Dict[str, Any]:
     message = str(exc)
     lower = message.lower()
 
+    connection_failed = any(
+        marker in lower
+        for marker in (
+            "cannot connect to",
+            "connection refused",
+            "all connection attempts failed",
+            "cannot reach embedding api",
+        )
+    )
+    timed_out = "timed out" in lower and any(marker in lower for marker in ("embedding", "ollama"))
+    if connection_failed or timed_out:
+        return {
+            "query": query,
+            "answer": (
+                "Knowledge retrieval failed because the configured embedding "
+                "service is unavailable. Check the embedding endpoint and "
+                "provider, start the service if it is local, then retry. "
+                f"Details: {message}"
+            ),
+            "content": "",
+            "provider": "llamaindex",
+            "error": message,
+            "error_type": "embedding_connectivity",
+            "log_message": "Embedding service unavailable during RAG query.",
+        }
+
     if "embedding provider returned invalid" in lower:
         return {
             "query": query,
--- a/deeptutor/tools/builtin/__init__.py
+++ b/deeptutor/tools/builtin/__init__.py
@@ -124,10 +124,13 @@ class RAGTool(_PromptHintsMixin, BaseTool):
             **extra_kwargs,
         )
         content = result.get("answer") or result.get("content", "")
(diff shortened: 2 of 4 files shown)


pancacake:
Half of this is already on `dev`, and the other half deserves its own PR. The messaging half — a tool failure reaching the model as its actual cause instead of "An unknown error occurred" — landed as `tool_error_message_factory(tool, error)` with the cause threaded through five call sites and all six i18n templates gaining an `{error}` placeholder. That also covers #1356: the Ollama adapter already raises a `ConnectionError` naming the endpoint and the remedy, and the dispatcher already catches it — the text was being discarded one layer up, in `pipeline.py`, not at the adapter. What is genuinely yours and not covered is `terminate_turn` for an unreachable embedding service (retrying cannot succeed, so the turn should end on the diagnosis). A PR doing only that would be welcome.

Issue #1347, opened 2026-09-10
[Bug]: read_skill conflates missing file with missing skill, seeding citation fabrication loop in chat


Sigebolo:
## Describe the bug

Two linked defects, reproduced end-to-end on v1.6.4 (multi-user, standard preset):

**Defect 1 — lying error.** `ReadSkillTool.execute(name="ncea-tutor", file="references/qbank_91164.md")` where the skill exists but the FILE does not returns `(skill not found: 'ncea-tutor' — ...)`. The name is correct; only the file is absent. The model reasonably concludes the *name* is wrong and starts renaming (`nccea-tutor` → `ncea_tutor` → …). The `ncea_tutor` variant then correctly fails the `^[a-z0-9][a-z0-9-]{0,63}$` check, which the model reads as further confirmation the name is the problem. Result in production: a 12s turn with 6+ wasted tool calls and **no answer delivered at all**.

**Defect 2 — fabricated Location after failed grounding.** When the model gives up on the bank and falls back to RAG, retrieved passages do not constrain composition: the turn issued 4 targeted `rag` calls (real schedule passages returned), then served a question cited as "2025 Exam, Question 3" about substances that appear in **no** 2025 paper text layer (verified by extracting the paper locally: real Q3 is diamond/graphite conductivity + ethanol/ZnO calculations + Lewis structures). A confident fake with a real-looking citation — twice, across model versions.

## Steps to reproduce (Defect 1)

1. Any skill dir with SKILL.md but without the requested reference file, e.g. skill `ncea-tutor`, file `references/qbank_91164.md`
2. `ReadSkillTool().execute(name="ncea-tutor", file="references/qbank_91164.md")`
3. Observe `(skill not found: 'ncea-tutor' …)` though the skill resolves fine

## Expected Behavior

1. Missing file in an existing skill must say so explicitly — name the skill as resolved, list its files, and forbid renaming (patch available, see below). Missing skill must list available skill names (one-round recovery instead of a guessing loop).
2. Related precedent already merged for other surfaces: #1115 (quiz evidence must match server-verified text), #1034 (only verified locators linkify). Chat question-serving has no equivalent gate — a `Location: <year> Q<part>` citation is currently unverified free text.

## Related Module

Skills / MCP

## Suggested fix (tested locally, 3 new tests green)

[...]

PR #1370, opened 2026-09-10
fix(skills): improve read_skill file recovery (#1347)

evan188199-tech:
## Summary
- complete the read_skill recovery scope after the core missing-file/missing-skill distinction landed in #1349
- list files in the resolved skill when the requested file is absent
- explicitly prevent fallback to other skill names for a file-path mistake
- list available skill names when the requested skill does not resolve
- add regression coverage for both recovery messages and the service-level missing-file path

## Testing
- [x] pytest focused recovery, service, registry, and scoped-resolution tests: 20 passed
- [x] pytest tests/services/skill tests/tools/builtin tests/runtime/registry tests/multi_user on the merged branch: 266 passed
- [x] ruff check and ruff format --check on changed files: passed
- [x] git diff --check and git diff --cached --check: passed

Builds on the distinction implemented by #1349 and addresses the remaining actionable recovery behavior in #1347. The broader citation-grounding defect remains out of scope.


evan188199-tech:
CI evidence after rebasing onto 30e0f7c3a: all Python import and Python test jobs pass. Web Node Tests fails in web/lib/chat-import/codex.ts:104 with `Property 'values' does not exist on type 'FileSystemDirectoryHandle'`. This PR has zero differences from origin/dev under web/, and the same frontend gate fails locally on the merged branch. The failure is therefore pre-existing on the current dev baseline and unrelated to this recovery-only change.

PR #1403, opened 2026-09-11
fix(mastery): bind draft before sending opening turn

### LINKED ISSUE #1392: [Bug]:New Mastery Path stays empty — the outline is written to a second, panel-invisible path

### Describe the bug

Severity: High — user-visible inconsistency + orphan duplicate records Component: Mastery Path — creation flow / conversation binding / panel listing Repro: 2 / 2

Summary
Creating a new Mastery Path and then designing its outline in a conversation produces two path records: an empty one that shows in the panel, and one holding the whole outline that never shows in the panel.

Steps to reproduce
In Mastery Path, create a new goal with a name.
On the "select learning materials" step, the picker is empty — nothing to select — so the step is skipped.
Open a new conversation to design the outline.
While the user is still typing, a pre-filled opening message is auto-submitted (the user did not send it). The assistant designs an outline and saves it to a second, conversation-owned path.
Return to the Mastery Path panel.
Expected
One goal = one path record. The outline is stored on the path the user created, the conversation is bound to it, and every path holding objectives is listed in the panel.

Actual
The path the user created stays empty (0 modules / 0 objectives) and shows 0% in the panel, even though an outline exists.
The full outline lives on a second path that never appears in the panel — it cannot be opened, renamed, or deleted through the UI.
Observable pattern: every panel-visible path id starts with topic_, every invisible one with unified_. Older unified_ paths do carry accumulated progress and may be visible, so the prefix is likely a symptom, not the cause — the failure looks specific to this creation + conversation flow.
The duplicate therefore cannot be removed by the user (panel-only operation) and the assistant has no delete-path tool either, so it persists as an orphan.
Suspected cause
Step 2 looks like the trigger: with an empty material picker, the newly created topic_ path is never wired up as the conversation's active path. The conversation falls back to its own unified_ session path, which is not registered in the index the panel reads from.

Suggested fixes
Bind the conversation to the path created in step 1 instead of creating a separate session path.
Do not auto-submit the pre-filled opening message.
Treat the empty material picker as an explicit state rather than continuing half-initialized.
List any path that holds objectives in the panel, or at least allow deleting/merging paths.
Workaround
Switching the conversation to the user-created path and rebuilding the outline there moves the work onto a visible record; the orphan record remains.

### Steps to reproduce

_No response_

### Expected Behavior

_No response_

### Related Module

Dashboard

### Configuration Used

_No response_

### Logs and screenshots

_No response_

### Additional Information

- DeepTutor Version:
- Operating System:
- Python Version:
- Node.js Version:
- Browser (if applicable):
- Related Issues:

KryptonGao:
### Description
Creating a Mastery Path could auto-submit its outline opening before React committed the new draft configuration. The first turn then omitted the intended Path ID, allowing the backend to write the outline to a separate, panel-invisible session-owned path.

Wait for the exact local draft key, workspace, Path ID, and session mode before allowing the opening or manual submissions. Keep server-session URL promotion separate so the first turn does not wait for the ID it creates. Ignore stale topic/session responses after navigation, and preserve the Mastery session mode when loading sessions and building turn requests.

Automatic openings and creation without materials remain supported. Historical orphan paths and backend fallback behavior are unchanged.

### Related Issues
Closes #1392

### Module(s) Affected
- [x] `web` (Frontend)
- [x] `tests`

### Validation
Rebased onto the latest `HKUDS/DeepTutor:dev` before validation.
- 25 rendered tests passed across 7 files, including 8 new regression cases using the real chat provider, session hook, and extracted opening hook with mocked API/transport.
- 88 targeted Node tests passed: Mastery, start-turn input, session loading, route selection, turn lifecycle/transport, and Watching session routing.
- Frontend typecheck, affected-file ESLint, repository hygiene, and `git diff --check` passed.
- Regression control: disabling the draft gate makes the first request lose its Path ID, workspace, and session mode; restoring it passes.

Coverage includes empty materials/modules, one opening under StrictMode, navigation races, failed and cancelled session loads, stored session modes, and URL promotion after the server binds the session. No live-backend end-to-end run was performed.

### Checklist
- [x] I have read and followed the contribution guidelines.
- [x] Documentation changes are not necessary.

Validation follows the repository's risk-based policy; the full pre-commit and regression suites were not run locally.

### DIFF
--- a/web/components/space/learning/MasteryStudy.tsx
+++ b/web/components/space/learning/MasteryStudy.tsx
@@ -31,6 +31,7 @@ import {
   useChatStateAdapter,
 } from "@/features/chat/ChatStateAdapter";
 import { useChatAutoScroll } from "@/hooks/useChatAutoScroll";
+import { useMasteryOpening } from "@/hooks/useMasteryOpening";
 import { useMasteryStudySession } from "@/hooks/useMasteryStudySession";
 import { useMeasuredHeight } from "@/hooks/useMeasuredHeight";
 import { useResearchOutlineContinuation } from "@/hooks/useResearchOutlineContinuation";
@@ -45,8 +46,6 @@ import { consumePendingPrompt } from "@/lib/pending-prompt";
 import { buildChatOutline, scrollToChatTurn } from "@/lib/chat-outline";
 import { buildConversationNotebookSave } from "@/lib/conversation-notebook-save";
 import {
-  MASTERY_OPENING_SCOPE,
-  masteryOpeningMessage,
   masterySessionRoute,
   type MasteryMode,
 } from "@/lib/mastery-mode";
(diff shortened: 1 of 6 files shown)


NoxsMedia:
Additional downstream evidence from official v1.6.8 (`a7c30a588f883ed7d92f806c4976c05b5b8fc2e1`): this also affects **Start review**, not just outline creation.

In three observed review launches, the generated opening naming the due items was persisted as a standalone ordinary chat. A separate, correctly bound Mastery Path conversation was created approximately 3, 4, and 12 seconds later and contains the subsequent review exchanges.

Read-only inspection of the persisted records showed:
- Each stray conversation contains exactly one user opening and one assistant response.
- Its first user-message request snapshot has `capability: "chat"`, empty tools/knowledge bases, and no `workspaceMode` or `masteryPathId`. Its stored workspace/path preferences are also empty.
- The corresponding Mastery conversations have `capability: "mastery_path"`, `workspaceMode: "mastery_path"`, and the intended `masteryPathId`; they contain 6, 12, and 22 messages respectively at inspection time.

The symptom is visible in the sidebar: the short opening chats sit outside the intended topic, while the actual review conversations are grouped inside it. This is consistent with the draft-binding race addressed here: the new-route loading condition does not wait for the configured draft to commit, and the opening sender can still read the previous ordinary-chat state.

This is evidence of the failure on the released baseline, **not validation of the fix**. I have not run PR head `f13012927f931c2250e0e0b946685850d2346ca7` against the live backend or reproduced it in a controlled browser test. No private conversation text, learner identifiers, or database attachments are included.

Issue #1406, opened 2026-09-11
[Feature Request]: Search across full conversation history


AoHanBei:
### Do you need to file a feature request?

- [x] I have searched the existing feature requests and pull requests.
- [x] I believe this is a legitimate feature request, not a question or bug.

### Feature Request Description

DeepTutor history pickers currently filter only the conversation title and the last message included in each session summary. A learner cannot find an older conversation by searching for a term that appeared earlier in the transcript.

Add bounded, server-side full-conversation search across user and assistant message text, with a short matching excerpt and a direct path back to the matching conversation.

### Proposed Scope

- Add a paginated session-history search API accepting a bounded literal query.
- Search session titles and persisted user/assistant message content, not tool traces or private provider metadata.
- Return one bounded best-match excerpt per session with role, message identifier, and timestamp where available.
- Preserve the existing user/workspace isolation and normal archived/imported-session visibility rules.
- Integrate the search into the existing history-session picker instead of loading every transcript in the browser.
- Debounce requests, cancel stale results, and retain the current title/last-message behavior for an empty query.
- Allow opening the matched session and locating the matching message when its identifier is available.
- Add tests for pagination, multiple matches in one session, literal `%`/`_` searches, empty queries, archived sessions, imported sessions, and cross-user isolation.

A first version does not need semantic/vector search. Literal case-insensitive matching keeps the behavior predictable and avoids adding an embedding dependency to session navigation.

### Related Module

Chat history / Session API

### Use Case

A learner remembers discussing `Bayes theorem` or a specific equation several weeks ago but does not remember the generated conversation title. Searching that phrase should find the conversation even when it is neither the title nor the final message.

### Additional Context

The current history picker already provides selection and lazy transcript preview. The missing piece is a bounded backend query over persisted messages and a result shape that includes a safe excerpt.

AoHanBei:
Claim: full-conversation-history-search

Planned scope:
- Add a bounded, paginated session search over titles and persisted user/assistant message text.
- Return one safe matching excerpt plus role/message metadata per session.
- Preserve user isolation and existing archived/imported-session visibility semantics.
- Wire the API into the history picker with debounce, stale-request cancellation, and message navigation.
- Cover pagination, literal wildcard characters, archived/imported histories, and cross-user isolation.

I am starting this as a focused feature PR targeting `HKUDS/DeepTutor:dev`. I found no existing Issue or PR covering this scope during the coordination check.

pancacake:
[...]

PR #1444, opened 2026-09-13
feat(sessions): soft-delete recycle bin with restore and permanent delete

sjagadee:
## Why

Users need a way to recover from accidental session deletions and to manage stale conversations without permanent data loss. The session store previously had no concept of deletion — once removed, a conversation was gone.

## What

Adds a **recycle-bin** to the session store. Every session operation that removes data now soft-deletes first, leaving the row intact for later recovery.

### Session soft-delete
- delete_session(id) — sets is_deleted=1 on the session row. The row and all messages remain in the database.
- restore_session(id) — clears is_deleted back to 0, recovering the conversation.
- hard_delete_session(id) — permanently removes the session row and all its messages. Requires a prior soft-delete (defence-in-depth).
- delete_session and restore_session are **idempotent**: calling them on an already-deleted or already-active session returns True without error.

### Query scope
All session-list and session-search operations (list_sessions, search_sessions) exclude soft-deleted sessions by default, consistent with existing include_deleted flags.

## How

- Added is_deleted INTEGER NOT NULL DEFAULT 0 column and deleted_at REAL column to the sessions table.
- Added composite index idx_sessions_is_deleted for efficient filtered queries.
- Added restore_session and hard_delete_session methods to SQLiteSessionStore.
- Updated all SQL queries to filter WHERE is_deleted = 0 by default.
- Added 25 unit tests covering soft-delete, idempotency, hard-delete guard, and restore.

## Tests

25 session-store unit tests — all passing.

Closes #1435

PR #1398, opened 2026-09-11
feat(sessions): add recycle bin for deleted chats

evan188199-tech:
Resolves #1389.

## Summary

Implements a recycle bin so deleted chat sessions are soft-deleted and can be restored or permanently purged instead of being irrecoverably removed.

## Backend

- Schema migration adds `deleted_at REAL` column to the `sessions` table (existing databases get the column via `ALTER TABLE`)
- `delete_session` now performs a soft delete (sets `deleted_at`; already-deleted sessions return `false`)
- Normal session lists (`_WHERE_NATIVE`, `_WHERE_IMPORTED`) exclude soft-deleted rows
- New store methods: `restore_session`, `purge_session`, `list_deleted_sessions`
- API endpoints:
  - `GET /sessions/recycle-bin` — list deleted sessions
  - `POST /sessions/{id}/restore` — clear the soft-delete marker
  - `DELETE /sessions/{id}/purge` — hard-delete plus attachment and mastery-path cleanup
  - `DELETE /sessions/{id}` now returns `recycled: true`

## Frontend

- `RecycleBinSection`: collapsible sidebar section listing deleted sessions with Restore and Delete permanently buttons
- Wired into `SidebarShell` via a new `recycleBinSlot` prop
- API client: `listRecycleBin()`, `restoreSession()`, `purgeSession()` with cache invalidation
- en/zh translations for all recycle-bin UI strings

## Verification

- `pytest tests/api/test_sessions_recycle.py` — 6 passed (soft-delete lifecycle, restore, purge, list exclusion)
- `pytest tests/tools/` — 221 passed, 2 skipped (regression)
- `npm run test:node` — 1131 passed
- `npm run typecheck` — PASS
- `npm run lint` — 0 errors, 70 pre-existing warnings
- `npm run i18n:check` — parity OK

PR #1445, opened 2026-09-13
feat(search): full conversation search with LIKE escaping

sjagadee:
## Why

Users need to search across their conversation history. The existing session list shows only session titles, with no way to find a conversation based on its actual content. Users with many sessions have no practical way to locate past discussions.

## What

Full conversation search using SQLite LIKE with proper escaping, excerpt highlighting, and ranked results.

### Search behaviour
- **LIKE search** across all message content within a session.
- **Ranked results**: sessions with a title match (message_id=0) are always listed first, before content matches.
- **Excerpts**: each matching session returns a text snippet with the matched term visible.
- **Pagination**: results are paginated (limit/offset) at the SQL level using ROW_NUMBER.
- **Character-safe escaping**: tilde (~) is used as the LIKE escape character — avoids Python string-literal backslash ambiguity that plagues ESCAPE '\'.

### Key SQL design
- _escape_like helper: replace('~', '~~').replace('%', '~%').replace('_', '~_')
- _SEARCH_SQL: ROW_NUMBER over a UNION ALL of title results and content results; title matches always ranked first via WHERE rn = 1 OR message_id = 0 ordering.
- Scalar subqueries for message_count and last_message — avoids LEFT JOIN filtering out sessions with no messages.
- NULL guard: if not rows or rows[0]['id'] is None: return [] — prevents phantom rows from edge-case empty UNION ALL.

## How

- Added search_sessions(query, limit, offset) method to SQLiteSessionStore.
- SQL query escapes special LIKE characters before building the LIKE pattern.
- Added 58 unit tests covering escaping, ranking, excerpts, pagination, and edge cases.

## Tests

58 session-store unit tests — all passing.

Closes #1436


sjagadee:
Closing as superseded. This branch conflicts with `dev` because full conversation search already shipped there in 00e2100a (#1415, `feat(chat): search full conversation history`, closes #1406) — merged just after this PR was opened.

Verified against current `dev`:
- `/api/sessions/search` + `search_sessions` implemented on **both** SQLite and PocketBase stores, plus a shared `deeptutor/services/session/search.py` helper (query normalization, excerpt bounding).
- Full frontend: debounced search in `HistorySessionPicker.tsx` with match highlighting/scroll-to, i18n (en/zh), OpenAPI contract + generated TS types — none of which existed in this PR.
- Dedicated coverage: `tests/api/test_session_search.py`, `tests/services/session/test_history_search.py`, plus PocketBase isolation tests.
- The LIKE-escaping problem this PR is built around doesn't exist on `dev`: message content is matched with `INSTR(LOWER(content), LOWER(?))`, a plain substring check, so `%`/`_` never need escaping in the first place. The tilde-escaped `LIKE` + `ROW_NUMBER` design here solves a bug class the shipped version structurally avoids.
- This branch's `protocol.py` also declares `search_sessions` (and, undocumented in the PR body, the recycle-bin methods from #1444) on `SessionStoreProtocol`, but only `sqlite_store.py` implements any of it — `pocketbase_store.py` is untouched, so any PocketBase deployment would hit `AttributeError` on search (same class of bug already fixed upstream for the recycle bin in #1398).

Nothing in this PR's diff adds capability missing from `dev`. No further conflict resolution needed on this branch.

PR #1446, opened 2026-09-13
feat: Persistent File Library with API router (issue #1437)

### LINKED ISSUE #1437: [Feature Request] Add a persistent reusable file library across conversations
### 问题与建议

DeepTutor 目前缺少一个可跨对话、跨学习流程复用的持久文件资料库。用户第一次上传某份资料后，之后在其他 Chat、精通之路、课程或其他学习流程中再次使用同一份资料时，仍需要重新上传，且可能需要重复解析。

这与“同一对话下一轮无法继续读取刚上传的附件”是不同问题：本 Issue 关注的是跨对话、跨学习场景的资料复用能力。

希望 DeepTutor 可以参考 ChatGPT 对已上传文件的处理方式：首次上传后，自动保存原始文件及其解析结果；之后用户再次使用同一份文件时，系统通过内容哈希或其他稳定标识识别已有资料，并直接复用已有文件和解析结果，而非要求重新上传或重新解析。

对于 MinerU 处理的本地资料，资料库条目应能关联原始文件、MinerU 生成的 Markdown、解析结果，以及 Markdown 引用的图片、表格和其他资源。这样同一资料以后可以在不同学习场景中继续使用。

建议默认持续保存、不自动过期；用户在需要时可以手动管理、替换、归档或删除资料。无需在每次上传时要求用户选择保存范围。

### 期望行为

1. 首次上传后自动建立可复用资料条目；
2. 在其他对话、精通之路、课程或学习流程中可再次选择和引用该条目；
3. 再次上传相同内容时自动复用已有文件与解析结果；
4. 对 MinerU 资料保留原始文件、Markdown 及相关资源的关联；
5. 默认不过期，提供后续手动管理入口；
6. 不强制将所有资料自动建立为全局知识库索引；资料复用与是否建立检索索引应是可分离的能力。

---

### Problem and proposal

DeepTutor currently lacks a persistent file library that can be reused across conversations and learning workflows. After a user uploads material once, using the same material later in another Chat, Mastery Path, course, or learning workflow still requires another upload and may require parsing again.

This is distinct from the problem where an attachment cannot be read in the next turn of the same conversation. This issue is about reusing materials across conversations and learning contexts.

DeepTutor could adopt an approach similar to ChatGPT’s handling of previously uploaded files: after the first upload, automatically preserve the original file and its parsed representation. When the same file is used again, identify the existing material through a content hash or another stable identifier and reuse the stored file and parsed result instead of requiring another upload or parse.

For local material processed by MinerU, a reusable library entry should be able to associate the original file, MinerU-generated Markdown, parsed representations, and images, tables, and other resources referenced by the Markdown. The same material could then be used again in different learning contexts.

The default should be persistent and non-expiring. Users can manage, replace, archive, or delete materials later. They should not have to choose a retention scope on every upload.

### Expected behavior

1. Automatically create a reusable material entry after the first upload;
2. Allow that entry to be selected and referenced from other chats, Mastery Paths, courses, and learning workflows;
3. Automatically reuse stored content and parsed results when identical content is uploaded again;
4. Preserve the relationship among original files, MinerU Markdown, and related resources;
5. Do not expire entries automatically by default, while providing later manual management;
6. Do not force every reusable material into a globally indexed knowledge base; material reuse and retrieval indexing should be separable capabilities.

sjagadee:
## Why

After a user uploads a file once, they have no way to reuse it in a different conversation or learning workflow — the same file must be uploaded and parsed again every time. This wastes bandwidth, parsing cost, and user time, especially for large materials that are used across multiple Mastery Paths, chats, or courses.

Users need a **persistent file library**: a reusable store of uploaded materials that survives individual conversations, identified by content hash so re-uploads are automatically deduplicated.

## What

A persistent file library backed by SQLite metadata and on-disk file storage, exposed via a REST API.

### FileLibraryStore (deeptutor/services/storage/file_library.py)

| Method | Behaviour |
|---|---|
| add_file(data, filename, mime_type) | Stores file to disk; returns existing entry if SHA-256 hash already exists (deduplication) |
| get_file(id) | Returns entry by id (includes soft-deleted) |
| delete_file(id) | Soft-delete (is_deleted=1), idempotent |
| hard_delete_file(id) | Permanent delete; requires prior soft-delete |
| restore_file(id) | Clears is_deleted; idempotent |
| list_files(limit, offset) | Active entries, newest-first, paginated |
| search_files(query) | Case-insensitive filename LIKE search |
| resolve_path(id) | Returns absolute Path or None |

### API Router (deeptutor/api/routers/file_library.py)

| Method | Path |
|---|---|
| GET | /files/library/ — list active entries |
| POST | /files/library/ — multipart upload with SHA-256 deduplication |
| GET | /files/library/search?q= — filename search |
| GET | /files/library/{id} — get entry |
| GET | /files/library/{id}/download — download file bytes |
| DELETE | /files/library/{id} — soft-delete |
| POST | /files/library/{id}/restore — restore |

### Deduplication design
- Content is identified by **SHA-256**. When add_file is called, the hash is computed and an active entry with the same hash is searched first.
- If found, the existing entry is returned without writing a new file to disk.
- If not found, the file is written atomically and a new DB row is created.

## How

- FileLibraryStore uses a SQLite table library_files with columns: id, sha256, filename, mime_type, size_bytes, library_path, created_at, updated_at, is_deleted, deleted_at.
- Files stored under data/user/workspace/library/files with UUID-based filenames.
- Atomic file writes via .tmp + os.replace.
- deeptutor/services/storage/__init__.py now exports FileLibraryStore and get_file_library_store.
- Router registered under /files/library with auth-gated endpoints.

## Tests

- 25 FileLibraryStore unit tests — all passing
- 20 API integration tests — all passing
- 6410 total tests pass (same 44 pre-existing failures unrelated to this change)

## Files

| File | Change |
|---|---|
| deeptutor/services/storage/file_library.py | New — full FileLibraryStore implementation |
| deeptutor/services/storage/__init__.py | Updated — exports FileLibraryStore, get_file_library_store |
| deeptutor/api/routers/file_library.py | New — REST API router |
| deeptutor/api/main.py | Updated — registered file_library router |
| tests/services/storage/test_file_library.py | New — 25 unit tests |
| tests/api/test_file_library_router.py | New — 20 API integration tests |

Closes #1437

### DIFF
--- a/deeptutor/api/main.py
+++ b/deeptutor/api/main.py
@@ -530,6 +530,7 @@ from deeptutor.api.routers import (
     tools as tools_router,
 )
 from deeptutor.api.routers.multi_user import router as multi_user_router  # noqa: E402
+from deeptutor.api.routers.file_library import router as file_library_router  # noqa: E402
 
 # Auth router is public — login/logout/register/status require no token
 app.include_router(auth.router, prefix="/api/auth", tags=["auth"])
@@ -559,7 +560,6 @@ app.include_router(
     tags=["multi-user"],
     dependencies=_auth,
 )
-
 app.include_router(question.router, prefix="/api/question", tags=["question"], dependencies=_auth)
 app.include_router(knowledge.router, prefix="/api", tags=["knowledge-bases"], dependencies=_auth)
 app.include_router(imports.router, prefix="/api/imports", tags=["imports"], dependencies=_auth)
@@ -570,6 +570,12 @@ app.include_router(
     mastery_path.router,
     prefix="/api/mastery-paths",
     tags=["mastery-path"],
+    dependencies=_auth,
+)
+app.include_router(
+    file_library_router,
+    prefix="/files/library",
+    tags=["library"],
     dependencies=_auth,
 )
 # WebSocket handlers authenticate inside the connection before ``accept``.
--- a/deeptutor/api/routers/file_library.py
+++ b/deeptutor/api/routers/file_library.py
@@ -1,0 +1,169 @@
+"""HTTP endpoints for the Persistent File Library (issue #1437).
+
+URL shape::
+
+    GET   /files/library/              — list files (newest first)
+    POST  /files/library/              — upload a file (multipart)
+    GET   /files/library/search        — search by filename (query param: q)
+    GET   /files/library/{file_id}     — get one entry
+    GET   /files/library/{file_id}/download  — download the file
+    DELETE /files/library/{file_id}     — soft-delete a file
+    POST  /files/library/{file_id}/restore    — restore a soft-deleted file
+
+The library is scoped to the authenticated user (via ``require_auth``) so
+each user sees only their own files.
+"""
+
+from __future__ import annotations
+
+import logging
+import mimetypes
+
+from fastapi import (
+    APIRouter,
+    Depends,
+    File,
+    Form,
+    HTTPException,
+    Query,
+)
+from fastapi.responses import FileResponse
+
+from deeptutor.api.routers.auth import require_auth
+from deeptutor.services.auth import TokenPayload
+from deeptutor.services.storage import file_library as _fl
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+def _get_store():
+    """Resolve the store at call time so tests can monkey-patch ``_fl``."""
+    return _fl.get_file_library_store()
+
+
+def _mime_type(filename: str) -> str:
+    mt, _ = mimetypes.guess_type(filename)
+    return mt or "application/octet-stream"
+
+
+# ── endpoints ──────────────────────────────────────────────────────────────
+
+
+@router.get("/", operation_id="library_list")
+async def list_library_files(
+    limit: int = Query(default=50, ge=1, le=200),
+    offset: int = Query(default=0, ge=0),
+    _auth: TokenPayload = Depends(require_auth),
+) -> list[dict]:
+    """List active library entries, newest first."""
+    return await _get_store().list_files(limit=limit, offset=offset)
+
+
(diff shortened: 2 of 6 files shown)


sjagadee:
### Code review

Found 1 issue:

1. `FileLibraryStore` singleton is cached under a hardcoded key instead of the resolved per-user root, so in a multi-user deployment every user after the first shares one library (files, listing, download, delete, restore all served from whichever user's request happened to warm the process). `get_file_library_store()` resolves `get_path_service().get_user_root()` (contextvar-scoped per authenticated user, see `deeptutor/multi_user/paths.py:get_current_path_service`) only once, then caches under the constant `key = "default"`:

https://github.com/HKUDS/DeepTutor/blob/01531261ee549dddad353b960d8f3e5058d14fcd/deeptutor/services/storage/file_library.py#L423-L436

This directly contradicts the router's own docstring, added in this PR, which promises per-user isolation:

https://github.com/HKUDS/DeepTutor/blob/01531261ee549dddad353b960d8f3e5058d14fcd/deeptutor/api/routers/file_library.py#L12-L15

The sibling store in the same package gets this right by keying the cache off the resolved root itself, so it should be mirrored here:

https://github.com/HKUDS/DeepTutor/blob/01531261ee549dddad353b960d8f3e5058d14fcd/deeptutor/services/storage/attachment_store.py#L234-L244

None of the router endpoints pass `_auth.user_id` into store resolution either, so fixing the cache key (e.g. `key = str(user_root)`, recomputed every call, matching `get_attachment_store`) is required to close the cross-tenant leak.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #1435, opened 2026-09-12
[Bug] v1.6.7 精通之路的首段等待无计时，且 RAG 工具事件协程未 await


Suyutao:
## 环境

- DeepTutor 1.6.7（PyPI 安装）
- Raspberry Pi（ARM64 / Debian 13 / Python 3.13）
- 以 systemd 运行，局域网 Web Dashboard

## 现象

在 Dashboard 的「精通之路」中请求制定/编辑学习大纲时，点击提交后会长时间停留在“DeepTutor 推理中…”。实际等待约 1–2 分钟后，才开始显示推理内容；此时思考计时器却从 `0:00` 开始，而不是累计此前已经等待的时间。

同一交互时间点，后端日志出现：

```text
RuntimeWarning: coroutine 'RAGService._emit_tool_event' was never awaited
```

这使用户无法判断系统是在准备学习路径、检索知识库、等待模型首包，还是已经失去响应。若后续模型流终于开始，UI 才出现思考计时，因而实际等待被隐藏。

## 观察与对照

- 发生时 `GET /api/mastery-paths/topics` 约 20 ms，`GET /api/mastery-paths/progress` 约 15 ms。
- 局域网和本机直连 `ws://.../ws/mastery-paths` 均可成功订阅。
- mastery SQLite 完整性检查正常，当前数据量很小（1 条路径、3 个事件）。
- Dashboard 静态页面与同源代理的 `/api/mastery-paths/topics` 都正常返回。
- 树莓派没有内存耗尽或 CPU 饱和。

因此当前更像是 RAG/精通之路的事件发送或阶段状态没有正确推进，而不是基础 API、WebSocket、数据库或局域网不可达。

## 另一个启动相关现象

重启已有用户工作区后，后端完成初始化有时超过启动器固定的 60 秒。一次观测中，从后端进程创建到 `Application startup complete` 约 78 秒；但启动器会在 60 秒时抛出“后端在 60 秒内未就绪”并终止仍在初始化的子进程，随后触发 systemd 重试。ARM 设备或已有工作区上这会造成不必要的服务短暂不可用。

## 期望

1. 确保 `RAGService._emit_tool_event` 的所有调用路径都被正确 `await` 或调度，避免丢失实时状态事件。
2. 从用户提交请求时开始展示总耗时，并明确区分“准备上下文 / 检索资料 / 等待模型响应 / 推理中”等阶段；不要只从首个推理事件开始计时。
[...]

PR #1449, opened 2026-09-13
fix(logging): schedule log coroutines from worker threads; make backend readiness budget configurable (#1435)

ZQR1101:
## Summary

Fixes two deterministic server-side halves of #1435 (the ARM/Raspberry Pi report). The other reported behaviors — a per-stage timer that starts at submission, and the "cannot connect" banner on a fresh mastery path — need either frontend work or more data from the reporter and are not in this PR.

**1. Log coroutines logged from worker threads were dropped, not scheduled.** `capture_process_logs` wraps the root logger with `ProcessLogHandler`, whose `emit` returned silently whenever the record arrived from a thread with **no running event loop** (`except RuntimeError: return`) — after `emit` had already created the coroutine. RAG retrieval logs precisely from such threads (ingestion, executors, `run_in_executor` bodies), so on every retrieval each worker-thread log line leaked a `RuntimeWarning: coroutine 'RAGService._emit_tool_event' was never awaited` — the exact warning in the report — and its raw-log event never reached the turn stream. That is a large part of why a running mastery outline shows an undifferentiated "DeepTutor 推理中…" with no live progress: the events that would describe the stages never arrive.

The handler now takes the loop the capture was created on and schedules thread-emitted coroutines back onto it with `asyncio.run_coroutine_threadsafe` — the same pattern the sibling `_NamedRawLogHandler` (LightRAG/GraphRAG loggers) in `RAGService._capture_raw_logs` already uses. Records logged on the loop thread keep the existing `ensure_future` path.

**2. The launcher's backend-ready budget is configurable.** `BACKEND_READY_TIMEOUT` was a hard-coded 60 seconds; the reporter measured 78 s from backend process start to `Application startup complete` on a Raspberry Pi with an existing workspace — so the launcher killed a healthy, still-initializing backend and systemd restarted it in a loop. `system.json` now accepts `backend_ready_timeout_s` (clamped to 30–600, default 60), consumed by the backend `_wait_for_http` wait. The `start.not_ready` message already interpolates the timeout, so the operator sees the configured budget in the error.

## Validation

Covered by the automated suite (confirmed red on the pre-fix code and green after):

- [x] A log record emitted from a worker thread while a capture with an async sink is active delivers its event onto the capture's loop — `test_capture_process_logs_schedules_coroutines_logged_off_loop`
- [x] `backend_ready_timeout_s` is clamped (30–600) and defaults to 60 — `test_system_settings_clamp_backend_ready_timeout`
- [x] The launcher helper reads the setting, falls back to the historical default, and clamps — `test_backend_ready_timeout_uses_system_setting`
- [x] Same-batch comparison against a pristine `dev` worktree (`tests/logging tests/runtime tests/services/config tests/api`): failure sets are the same pre-existing environment debt plus one known cross-file `WorkspaceError` singleton leak that fails 8 unified-ws tests on pristine and 10 on the changed tree — the +2 are this PR's new turn-level tests joining that already-polluted sensitive set, not new breakage (both pass in isolation and in the narrower batches); `ruff check` / `ruff format --check` (0.16.0) clean

Not verified — needs the reporter's environment:

- [ ] On-device confirmation that the mastery outline turn now shows live progress events and that `system.json: {"backend_ready_timeout_s": 120}` lets an 80-second initialization complete


ZQR1101:
Superseded by upstream's own fixes, which landed while this PR was in review — closing in favor of them:

- `df161577` (`fix(logs): deliver a retrieval progress event logged off the event loop`) fixes the same `ProcessLogHandler` drop this PR did: the loop captured at handler construction, worker-thread records scheduled back with `run_coroutine_threadsafe`, and — nicer than my version — an explicit `coro.close()` for the case where the capture outlives its loop. Same files, same issue reference, plus the RAG service side.
- `883c90f2` (`fix(launcher): stop killing a backend that is still starting`) makes the readiness budget overridable (`DEEPTUTOR_BACKEND_READY_TIMEOUT` / `DEEPTUTOR_FRONTEND_READY_TIMEOUT`), citing the same 78-second ARM measurement. Environment variables are the better mechanism for a systemd deployment than my system.json key, so nothing of this PR's half is lost.

My thanks to the maintainer for moving fast on both. Verified upstream's versions against my diagnosis before closing: the worker-thread drop (#1435's RuntimeWarning + lost retrieval progress) and the 60-second kill are both covered. The issue's remaining halves — the timer that starts at the first reasoning event, per-stage progress, and the fresh-path connection banner — remain open on #1435.

PR #1463, opened 2026-09-14
50/50 RevShare Integration: DeepTutor & AIML API

hugoaimlapi:
Hi! I'm Hugo from aimlapi.com — an AI aggregator that gives access to 1000+ models in one API, trusted by 400k+ users.

We'd love to be available as a verified provider option inside DeepTutor — so we went ahead and did all the technical work on our side.

To build our partnership, we offer a 50/50 revenue share on all traffic from this integration. (P.S.: tracking starts as soon as this release goes live, so no earnings will be lost during setup)

My contacts: <hugo@aimlapi.com> (email / Slack), Telegram: @hug0the

---

### Description

Today DeepTutor can reach AI/ML API only through a hand-typed `custom` profile: the endpoint, the env var and the capability flags all have to be rediscovered by each user, and nothing in the registry tells the config loader that an `https://api.aimlapi.com/v1` base belongs to a gateway that routes any model. And unlike OpenRouter, those requests carry no attribution.

Four commits, ~250 lines, no new dependency:

**1. `feat: add aimlapi.com as an OpenAI-compatible gateway provider`**

- `deeptutor/services/provider_registry.py` — one `ProviderSpec`: `name="aimlapi"`, `display_name="aimlapi.com"`, `env_key="AIMLAPI_API_KEY"`, `backend="openai_compat"`, `is_gateway=True`, `detect_by_base_keyword="aimlapi"`, `default_api_base="https://api.aimlapi.com/v1"`, in the same gateway block as the other aggregators; plus `aiml` / `aiml_api` in `PROVIDER_ALIASES`.
  No `detect_by_key_prefix` — AI/ML API keys carry no distinguishing prefix, so the endpoint is the only signal that will not misfire on someone else's key. No `strip_model_prefix` — model ids keep their vendor prefix (`openai/gpt-4o-mini`), unlike aihubmix.
- `deeptutor/services/llm/capabilities.py` — an `aimlapi` entry carrying the same flags as the neighbouring aggregators: `supports_response_format`, `supports_streaming`, `supports_tools`, `supports_vision`, `system_in_messages`.
- `deeptutor_cli/init_wizard.py` — four `LLM_FALLBACK_MODELS` ids. Each was checked against `GET https://api.aimlapi.com/v1/models`, filtered to `type == "openai/chat-completions"` and matched against both ids and aliases, rather than copied from another aggregator's list.

**2. `feat: attribute DeepTutor traffic to aimlapi.com by endpoint host`**

- `deeptutor/services/llm/openai_http_client.py` — `AIMLAPI_ATTRIBUTION_HEADERS` (`HTTP-Referer`, `X-Title`, `X-AIMLAPI-Partner-ID`, `X-AIMLAPI-Source`) merged into the per-call header dict in `openai_sdk_client_kwargs()`, in exactly the place `OPENROUTER_ATTRIBUTION_HEADERS` already occupies. `HTTP-Referer` / `X-Title` identify DeepTutor as the calling app — they are not AI/ML API's own URL and title.
- The scoping is deliberately **narrower** than the OpenRouter check next to it. That one matches a substring of the URL, which also fires for a proxy whose path merely contains the vendor name. `_endpoint_host()` + `_uses_aimlapi()` compare the *host* of the resolved endpoint against an exact allow-list, so the headers cannot ride a request to `api.aimlapi.com.evil.io`, to `notaimlapi.com`, or to a self-hosted proxy configured under an `"aimlapi"` binding.
- Headers are merged into a per-call dict rather than assigned, so a profile's own `extra_headers` still win and the module constant is never mutated.

**3. `fix: drop repeated ids from a provider's /models list`** — not aimlapi-specific

- `deeptutor/services/llm/utils.py` — `collect_model_names()` keeps the first occurrence of each id. A `/models` payload may list one model id once per endpoint family it serves, so the same name arrives more than once: AI/ML API returned 941 rows for 789 distinct ids on 2026-09-14. Every repeat became a separate, identical row in the model picker with no way for the user to tell them apart. First occurrence wins, so payload order — which providers use to put their preferred models first — is preserved.

**4. `fix(aimlapi): use the registered partner id`** — swaps the pre-registration placeholder for the id the gateway actually minted.

### Related Issues

- None — opening this directly. Happy to file a tracking issue first if you prefer.

### Module(s) Affected
- [ ] `agents`
- [ ] `api`
- [ ] `config`
- [ ] `core`
- [ ] `knowledge`
- [ ] `logging`
- [x] `services`
- [ ] `tools`
- [ ] `utils`
- [ ] `web` (Frontend)
- [ ] `docs` (Documentation)
- [ ] `scripts`
- [x] `tests`
- [x] Other: `deeptutor_cli` (wizard fallback model list)

### Checklist
- [x] **This PR targets `dev`** (not `main` — see the base-branch selector above).
- [ ] I have updated the documentation (if necessary) — no doc change needed: the wizard and the web model picker are both driven by `PROVIDERS`.

### Additional Notes

**How to test**

```bash
pytest -q tests/services/test_provider_registry.py \
          tests/services/llm/test_openai_http_client.py \
          tests/services/llm/test_utils.py
```

Attribution is scoped to the host, not to the binding:

```bash
python -c "
from deeptutor.services.llm.openai_http_client import openai_sdk_client_kwargs
from deeptutor.services.provider_registry import find_by_name
spec = find_by_name('aimlapi')
for base in ('https://api.aimlapi.com/v1', 'https://proxy.example.com/v1'):
    h = openai_sdk_client_kwargs(api_key='sk-test', base_url=base, spec=spec)['default_headers']
    print(base, '->', {k: v for k, v in h.items() if k.startswith('X-AIMLAPI-')})
"
# https://api.aimlapi.com/v1   -> {'X-AIMLAPI-Partner-ID': 'part_...', 'X-AIMLAPI-Source': 'agent/deeptutor'}
# https://proxy.example.com/v1 -> {}
```

Live run: `export AIMLAPI_API_KEY=<key>`, then `deeptutor init` → pick **aimlapi.com** → the endpoint resolves to `https://api.aimlapi.com/v1` and the model list arrives de-duplicated.

**Notes**

- No UI change. `web/lib/llm-options.ts` takes the provider label from the backend registry, so the new binding appears on its own. There is no vendored logo for aimlapi under `web/public/provider-icons/`, so `ProviderIcon` falls back to the generic mark — same as atlascloud, novita and orcarouter today. Happy to add an SVG as a follow-up if you'd like one.
- Nothing in routing, the agent loop or the config loader changes; previously saved profiles load unchanged. The provider is appended to the existing gateway block — no reordering, no change to which provider the wizard preselects.


deeptutor/services/llm/openai_http_client.py
```diff
@@ -25,6 +26,23 @@
...
+    "X-AIMLAPI-Partner-ID": "part_ItAs0L5uSTvV2dFDOZZaS1BL",
+    "X-AIMLAPI-Source": "agent/deeptutor",
+}
+
+# Exact hosts these headers may be sent to. Matching the *host* of the resolved
+# endpoint — not a substring of the URL and not the configured provider name —
+# is what keeps attribution off a look-alike domain ("api.aimlapi.com.evil.io",
+# "notaimlapi.com") and off a self-hosted proxy that merely fronts the same API
+# under a binding still typed as "aimlapi".
+_AIMLAPI_ATTRIBUTION_HOSTS: frozenset[str] = frozenset({"api.aimlapi.com"})
```
thelooter: This feels very LLM-prose-y. Might wanna do another pass at those comments to condense them down to whats needed. And ideally no em-dashes :)

hugoaimlapi: thanks for the feedback mate! yeah, we use LLMs to add more details, but there's a lot of text :) we'll reduce it

thelooter:
Found a case where the partner headers get sent to the wrong host.

`_endpoint_host()` falls back to `spec.default_api_base` when there's no base_url, but `_build_openai_client` passes `base_url=config.base_url or None` to AsyncOpenAI, so the SDK uses its own default:

https://github.com/HKUDS/DeepTutor/blob/04edf444fc2c5ec3e792ddfab1bc10a8a304e00d/deeptutor/services/llm/openai_http_client.py#L189
https://github.com/HKUDS/DeepTutor/blob/04edf444fc2c5ec3e792ddfab1bc10a8a304e00d/deeptutor/runtime/agentic/client.py#L169

```python
_build_openai_client(
    LLMClientConfig(binding="aimlapi", model="openai/gpt-4o-mini", api_key="k", base_url=None),
    disable_ssl_verify=False,
)
# base_url: https://api.openai.com/v1/
# headers: HTTP-Referer, X-Title, X-AIMLAPI-Partner-ID, X-AIMLAPI-Source
```

The settings resolver fills in the default URL first, so the normal path is fine. Anything that builds LLMClientConfig directly runs into it though. The check should use the URL the client actually calls. `test_aimlapi_attribution_sent_for_the_registry_endpoint` also passes `base_url=None`, so right now it tests the wrong behavior.

thelooter:
The de-dup makes sense, but the list still isn't great for picking an LLM. `/v1/models` right now returns 943 rows with 791 distinct ids, and only 359 of those are `openai/chat-completions`. Video, image, TTS, STT and embedding models all still show up in the picker and only fail once you try to chat with them. Filtering on `type` would fix that. Fine as a follow-up.

thelooter:
Two smaller things:

- atlascloud and novita also have tests in `tests/services/config/test_provider_runtime.py` (default endpoint + base URL detection), `tests/api/test_settings_router.py` (label/base_url) and an entry in `test_registered_cloud_openai_compat_providers_enable_native_tools`. Would be good to have the same for aimlapi.
- The attribution check is an exact host match, but `_uses_openrouter` right above it is still a substring match, and `detect_by_base_keyword="aimlapi"` matches any URL containing "aimlapi" (e.g. `https://proxy.corp/aimlapi/v1`). Not blocking, just confusing to have both approaches next to each other when someone adds the next gateway.

thelooter:
Question for the maintainers: the partner ID (`part_ItAs0L5uSTvV2dFDOZZaS1BL`) is hardcoded and was issued by AI/ML API. Is it registered to an HKUDS account? If not, it's probably better to have one issued to you or read it from config, since changing it later means a new release.

deeptutor/services/llm/openai_http_client.py
```diff
@@ -138,6 +186,8 @@ def openai_sdk_client_kwargs(
         headers["x-session-affinity"] = uuid.uuid4().hex
     if _uses_openrouter(spec, base_url):
         headers.update(OPENROUTER_ATTRIBUTION_HEADERS)
+    if _uses_aimlapi(spec, base_url):
```
thelooter: This can send the partner headers to the wrong host. `_endpoint_host()` falls back to `spec.default_api_base` when there's no base_url, but `_build_openai_client` passes `base_url=config.base_url or None` to AsyncOpenAI, so the SDK uses its own default:

https://github.com/HKUDS/DeepTutor/blob/04edf444fc2c5ec3e792ddfab1bc10a8a304e00d/deeptutor/runtime/agentic/client.py#L169

```python
_build_openai_client(
    LLMClientConfig(binding="aimlapi", model="openai/gpt-4o-mini", api_key="k", base_url=None),
    disable_ssl_verify=False,
)
# base_url: https://api.openai.com/v1/
# headers: HTTP-Referer, X-Title, X-AIMLAPI-Partner-ID, X-AIMLAPI-Source
```

The settings resolver fills in the default URL first, so the normal path is fine. Anything that builds LLMClientConfig directly runs into it though. The check should use the URL the client actually calls. `test_aimlapi_attribution_sent_for_the_registry_endpoint` also passes `base_url=None`, so right now it tests the wrong behavior.

deeptutor/services/llm/utils.py
```diff
@@ -323,11 +323,25 @@ def _normalize_model_name(entry: object) -> str | None:
...
+    shows the repeats as separate, identical choices.
+
+    Args:
+        entries: Raw provider payload entries, each a mapping or a bare string.
+
+    Returns:
+        Model names in payload order, without repeats.
+    """
     names: list[str] = []
+    seen: set[str] = set()
```
thelooter, line 349: The de-dup makes sense, but the list still isn't great for picking an LLM. `/v1/models` right now returns 943 rows with 791 distinct ids, and only 359 of those are `openai/chat-completions`. Video, image, TTS, STT and embedding models all still show up in the picker and only fail once you try to chat with them. Filtering on `type` would fix that. Fine as a follow-up.

tests/services/test_provider_registry.py
```diff
@@ -53,6 +53,33 @@ def test_novita_provider_aliases_and_base_detection() -> None:
     assert find_gateway(api_base="https://api.novita.ai/openai") == spec
 
 
+def test_aimlapi_provider_aliases_and_base_detection() -> None:
```
thelooter, line 56: atlascloud and novita also have tests in `tests/services/config/test_provider_runtime.py` (default endpoint + base URL detection), `tests/api/test_settings_router.py` (label/base_url) and an entry in `test_registered_cloud_openai_compat_providers_enable_native_tools`. Would be good to have the same for aimlapi.

deeptutor/services/llm/openai_http_client.py
```diff
@@ -114,6 +132,36 @@ def _uses_openrouter(spec: "ProviderSpec | None", api_base: str | None) -> bool:
...
+    # as a path; "//" makes it a netloc without guessing http vs https.
+    if "//" not in resolved:
+        resolved = f"//{resolved}"
+    try:
+        return (urlsplit(resolved).hostname or "").lower()
+    except ValueError:
+        return ""
+
+
+def _uses_aimlapi(spec: "ProviderSpec | None", api_base: str | None) -> bool:
```
thelooter: This is an exact host match, but `_uses_openrouter` right above is still a substring match, and `detect_by_base_keyword="aimlapi"` in the registry matches any URL containing "aimlapi" (e.g. `https://proxy.corp/aimlapi/v1`). Not blocking, just confusing to have both approaches next to each other when someone adds the next gateway.

deeptutor/services/llm/openai_http_client.py
```diff
@@ -25,6 +26,23 @@
     "X-OpenRouter-Title": "DeepTutor",
 }
 
+# AI/ML API attributes traffic the same way OpenRouter does, plus two headers
+# of its own. HTTP-Referer / X-Title identify DeepTutor as the calling app —
+# they are not AI/ML API's own URL and title.
+AIMLAPI_ATTRIBUTION_HEADERS: dict[str, str] = {
+    "HTTP-Referer": "https://github.com/HKUDS/DeepTutor",
+    "X-Title": "DeepTutor",
+    "X-AIMLAPI-Partner-ID": "part_ItAs0L5uSTvV2dFDOZZaS1BL",
```
thelooter, line 33: Question for the maintainers: this partner ID is hardcoded and was issued by AI/ML API. Is it registered to an HKUDS account? If not, it's probably better to have one issued to you or read it from config, since changing it later means a new release.

Issue #1471, opened 2026-09-15
[Feature Request]: Adaptive long-term learning engine for Mastery Path — retention-based spaced repetition, unified learning evidence, and study prioritization


yepyhun:
### Do you need to file a feature request?

- [x] I have searched the existing feature request and this feature request is not already filed.
- [x] I believe this is a legitimate feature request, not just a question or bug.

### Feature Request Description

Hi,

I reviewed the current implementation in `main`, not just the README. I looked through the relevant learning, mastery, scheduling, service, and course-state code.

DeepTutor's current Mastery Path already has a strong foundation:

- deterministic per-objective mastery gates
- persisted quiz attempts and errors
- qualitative assessment for concepts/design tasks
- a spaced-review queue
- `next_objective()` prioritizing due review before new material
- Course-level aggregation of Mastery Paths, Question Bank data, syllabus units, and reading workspaces

The biggest missing piece, in my opinion, is not "adding spaced repetition", because DeepTutor already has it.

The bigger opportunity is turning the current fixed-interval review system into a genuinely adaptive long-term learning engine.

The most important part of this request is adaptive spaced repetition. Cross-course study planning is useful too, but secondary.

CURRENT BEHAVIOR

1. Spaced repetition currently uses predefined intervals

In `deeptutor/learning/scheduler.py`, the intervals are currently defined roughly as:

MEMORY:
[0, 1, 3, 7, 14, 30, 60]

CONCEPT:
[3, 7, 14, 30]

PROCEDURE:
[3, 7, 14]

DESIGN:
[14, 28]

`schedule_next()` moves forward or backward through these lists depending on correctness / consecutive correct answers.

This is already useful, but it means that learners with very different memory characteristics still mostly follow the same interval structure.

2. RepetitionState does not model actual memory state

In `deeptutor/learning/models.py`, `RepetitionState` currently stores things like:

- interval_index
- consecutive_correct
- consecutive_wrong
- next_review_at

What is missing is a richer model of long-term memory, for example:

- difficulty
- stability
- retrievability / estimated probability of recall
- desired retention
- lapse count
- review count
- last review timestamp
- uncertainty/confidence in the estimate

This matters because the current system can answer:

"Which interval index is this item on?"

but not really:

"How likely is this learner to remember this knowledge point tomorrow, in one week, or by exam day?"

3. Mastery and retention should be separated

[...]

PR #1475, opened 2026-09-15
feat: complete adaptive SRS persistence and review UI

KryptonGao:
## What changed

Completes the Adaptive SRS MVP for issue #1471 across SQLite, API projections, and the existing Mastery Path UI.

- Adds the `mastery_learning_evidence` SQLite projection with migration/backfill, indexed queries, cascade cleanup, and transactional synchronization with `state_json`.
- Exposes fixed-time read-only review projections with forgetting risk, scheduling reasons, memory metrics, recent failures, and evidence reports.
- Updates the Mastery Path UI with due/future review ordering, risk/reason explanations, expandable objective memory details, retention evidence, responsive layouts, and direct review routing.
- Preserves legacy JSON compatibility and existing idempotency behavior.

## Validation

- `.venv/bin/python -m pytest deeptutor/learning/tests -q` — 514 passed
- `.venv/bin/python -m pytest deeptutor/learning/tests tests/capabilities/test_mastery_loop_pipeline.py tests/capabilities/test_mastery_capability.py tests/capabilities/test_mastery_echoed_options.py -q` — 551 passed
- `npm run typecheck`
- `npm run lint -- --quiet`
- `npm run test:unit` — 74 tests passed
- `npm run contracts:check`
- `npm run i18n:check`
- SQLite projection smoke test
- Browser acceptance on desktop and 390×844 mobile viewport, including review routing, detail/evidence rendering, internal scrolling, no horizontal overflow, and persistence after navigation/reload


yepyhun:
Thanks for implementing this. The backward-compatible state extension, transactional persistence, and review explanations are useful foundations for #1471. After reviewing the current code at `96b6cb89`, I have a few follow-up points that would make the retention behavior and long-term persistence more reliable.

These observations come from static code review; I have not run the PR's test suite. I also appreciate that this is explicitly described as a cold-start baseline, not a calibrated FSRS implementation.

### 1. Let review timing influence the memory update

In [`schedule_review()`](https://github.com/HKUDS/DeepTutor/blob/96b6cb89d3c838735bef2860b4d7fa7fb341d77e/deeptutor/learning/scheduler.py#L173-L214), successful retrieval multiplies stability by `1.2 + quality * 1.5`, with a consecutive-success adjustment. The update does not use elapsed time or pre-review retrievability, and difficulty does not participate in that calculation.

Consequently, with the same starting state and answer quality, a successful answer one minute later produces the same stability gain as one a month later. Repeated immediate practice can therefore extend the interval without demonstrating long-term retention.

Suggested direction: use elapsed time/pre-review recall in the update, with explicit handling for same-session repetitions. An established scheduler implementation is one option; a documented conservative baseline is also reasonable for this phase. Either way, distinguish repeated practice from successful delayed retrieval, and present recall percentages as model estimates rather than validated personal probabilities.

### 2. Make evidence replay reproduce the actual service behavior

[`record_qualitative_in_memory()`](https://github.com/HKUDS/DeepTutor/blob/96b6cb89d3c838735bef2860b4d7fa7fb341d77e/deeptutor/learning/service.py#L1118-L1151) always appends evidence, but the first successful assessment only initializes the repetition state. Existing states are updated only when already due. By contrast, [`replay()`](https://github.com/HKUDS/DeepTutor/blob/96b6cb89d3c838735bef2860b4d7fa7fb341d77e/deeptutor/learning/scheduler.py#L307-L319) schedules every recorded event.

Replaying a first successful qualitative assessment therefore produces a different review count/stability from the live service. Early assessments create a similar mismatch. Please use the same event-to-state transition for live updates and replay, or persist the information needed to reproduce deliberately skipped scheduling updates. Legacy migrated state also needs a clear replay starting point.

A particularly useful focused check would compare the actual service-produced state with replay of its persisted history, including initial and not-yet-due qualitative assessments.

### 3. Avoid rebuilding the complete evidence projection on ordinary writes

[`_sync_evidence_projection()`](https://github.com/HKUDS/DeepTutor/blob/96b6cb89d3c838735bef2860b4d7fa7fb341d77e/deeptutor/learning/storage.py#L626-L666) deletes every evidence row for a path and reinserts its entire history. Both ordinary saves and mutation commits call it, even when evidence has not changed.

This adds work proportional to the full history on each action, which is a poor fit for semester-long use. Keep the transaction guarantee, but update only added/changed evidence during normal operation; reserve full rebuilds for migration, reset, or explicit repair. This is a scaling concern visible in the code, not a claim that I measured current slowness.

### 4. Finish the configuration and scheduler integration

The `desired_retention` field exists, but I could not find a global/per-path configuration route; new states use the fixed `0.9` constant. Likewise, the new `RetentionScheduler` protocol is not used by the main consumers, which still construct/depend on `SpacedRepetitionScheduler` directly.

Please expose a validated retention setting and route scheduler construction through the existing owning layer. No larger plugin framework is needed. These would make the advertised configurability and replaceability usable in practice.

The line count itself is not my concern: a substantial portion is tests and necessary persistence/UI integration. I would prioritize consistent memory transitions, replayable evidence, and bounded write work before expanding the UI further. Full parameter calibration, misconceptions, and cross-course planning can reasonably remain follow-up work.

### Practical completion criteria for this phase

To make the requested outcome unambiguous, I would consider the adaptive-SRS foundation complete when the following observable cases hold:

- A failed review schedules an earlier retry, while successful retrievals separated in time can progressively extend the interval. Repeated answers in the same study session cannot produce an unjustifiably long interval.
- Given the same persisted starting state, settings, and ordered evidence events, replay produces the same stability, difficulty, review/lapse counts, and next-review time as the live service. This includes first assessments, qualitative assessments, not-yet-due assessments, and legacy migrated paths.
- Desired retention is validated, configurable at the documented scope, persists across restart, and actually changes scheduling. Explanations shown to the learner are derived from the same values used by the scheduler.
- Saving unrelated path state does not rewrite the complete evidence history. Normal evidence persistence scales with the new/changed events; explicit migration or repair may still rebuild the projection.
- An existing pre-migration Mastery Path can be loaded, reviewed, restarted, and loaded again without losing mastery, retention state, evidence, or queue ordering.

These do not require a large new framework. A small set of deterministic scheduler checks plus one persistence/restart flow would provide strong evidence. Once these criteria and the concerns above are met, I would regard the bounded Phase 1 adaptive-review foundation as complete; broader misconception tracking and cross-course planning can remain separate work.

PR #1480, opened 2026-09-15
fix(rag): stop a stalled LlamaIndex worker leaking progress into the next job (#1478)

### LINKED ISSUE #1478: LlamaIndex re-index can leak stale embedding progress after a timeout
## Summary

On Windows, a LlamaIndex knowledge-base re-index can time out while its synchronous embedding worker continues running in the background. The worker can then emit progress into a later task through shared global embedding settings, causing misleading repeated progress and leaving no queryable index.

## Environment

- DeepTutor v1.6.7 source checkout
- Windows 11
- RAG provider: `llamaindex`
- Embeddings: local Ollama, `qwen3-embedding:0.6b` (1024 dimensions)
- Re-index input: two PDFs, including one large textbook

## Reproduction

1. Create a LlamaIndex knowledge base with one or more large PDFs.
2. Configure a local Ollama embedding endpoint.
3. Start Re-index and leave the page open.
4. If an embedding call stops completing, wait for the indexing stall timeout, then observe the task/progress stream or retry the operation.

## Actual behavior

- The UI may repeatedly show an entire batch sequence, e.g. `Embedding batches: 1/103` through `103/103`, then restart from `1/103`.
- The process-log count continually increases even though it is not a count of independent workers.
- A stalled task can eventually report a timeout, but the underlying executor worker keeps running because synchronous Python code cannot be cancelled.
- Progress is attached through the process-global `Settings.embed_model` callback, so a stale worker can write progress intended for a newer task.
- The KB can remain in a processing-looking state even after reaching `103/103`, without a persisted `docstore.json` / queryable index.

## Expected behavior

- One re-index task should own one progress stream.
- Once a task is marked failed/cancelled, its worker must no longer update user-visible progress.
- A retry should not overlap a still-live worker or receive its callbacks.
- The UI should not append duplicate process/progress lines indefinitely.
- A KB should only be reported ready after the complete persisted index is present and loadable.

## Likely direction

A robust fix appears to be to create a per-index-operation embedding adapter/callback rather than mutating the global `Settings.embed_model` callback, detach that callback in `finally`, and prevent a new LlamaIndex indexing job from starting while a timed-out worker is still alive. The progress UI should de-duplicate interleaved raw process-log and structured progress events.

This is particularly visible on long local Ollama embedding jobs, where a retry can otherwise look like many parallel processes while producing no usable index.

ZQR1101:
## Summary

Fixes #1478: a LlamaIndex re-index that times out on a stalled embedding call left a synchronous worker running in a thread Python cannot interrupt — and that worker kept writing batch progress into the *next* job, so retries showed endlessly repeated/restarting "Embedding batches" progress and no usable index.

**Root cause.** The LlamaIndex adapter's progress callback lives on a process-global `Settings.embed_model` singleton with exactly one slot (`set_progress_callback`). The stall guard reads progress through a `_heartbeat` wrapper registered into that slot. When a run stalls, the guard raises `IndexingStallError`, but the sync embedding worker keeps running in its thread and keeps calling the stored callback. Nothing marked the run as "finished/aborted" to the progress path — so once a later job (or a manual retry) re-armed the slot, the stale worker's batch events flowed straight into the new job. Overlapping runs also displaced each other's heartbeat on every poll tick (the existing code dripped this in a comment).

**Fix.** The stall guard now stamps each run with a unique owner and the heartbeat wrapper refuses to forward progress once that run no longer owns the shared slot:

- On entry, the guard claims the slot (`_ACTIVE_PROGRESS_OWNER = <this run>`).
- `_heartbeat` returns early unless the slot still belongs to this run — so a stale worker from a finished/aborted run becomes a no-op.
- In a `finally`, the guard releases the slot **only if it still holds it**, so a newer run's claim is never clobbered.

The `set_progress_callback` signature is unchanged, so normal single-run indexing and the ordinary progress path are unaffected; the change only makes finished/stalled runs stop writing into the global slot.

## Validation

Covered by the automated suite (confirmed red on the pre-fix code and green after):

- [x] A worker that keeps emitting after its run finishes no longer writes progress — `test_finished_job_progress_callback_stops_leaking_into_next_turn`
- [x] A stale wrapper is a no-op once a newer run owns the slot (fail-closed on ownership handoff) — `test_stale_wrapper_from_finished_run_is_suppressed_after_ownership_changes`
- [x] Existing stall-guard behavior unchanged: stall raises, progress-keeps-flowing completes, a concurrent job displacing the slot doesn't false-stall, user progress is forwarded, and `initialize`/`add_documents` fail bounded — 7 prior tests `tests/services/rag/test_llamaindex_pipeline_stall.py`
- [x] Same-batch comparison against a pristine `dev` worktree (`tests/services/rag tests/api`): failure sets byte-identical (13 pre-existing environment failures); `ruff check` / `ruff format --check` (0.16.0) clean

Not verified — needs a real stalled embedding provider:

- [ ] On-device (Windows + local Ollama): a re-index that times out, then a retry, no longer shows repeated/reset "Embedding batches" progress, and the KB only reports ready with a persisted, loadable index


yepyhun:
On-device reproduction details: per-outer-group progress reaches 103/103 multiple times and is presented as total 100%; Windows suspend time is counted by the stall watchdog, causing a false stall immediately on resume while Ollama continues returning HTTP 200. Full evidence and proposed fixes will follow once GitHub accepts this comment.

yepyhun:
Full evidence from the same Windows + local Ollama reproduction:

### Misleading repeated 100%

The local adapter uses LlamaIndex `embed_batch_size=1024` and provider `batch_size=10`. Each outer group therefore reports `Embedding batches: 1/103 ... 103/103`. A larger book starts another outer group and resets the same counter, so this per-group denominator is incorrectly presented to the user as total job completion.

Ollama's server log recorded 300 successful `POST /api/embed` calls for the run: 209 before suspend and 91 after resume, with zero non-200 responses. This is consistent with multiple outer groups, not a provider hang. The user saw the second `103/103` as 100% complete, but another large group still remained.

Suggested fix: aggregate progress by total indexed nodes across outer calls. If the total is unavailable at that layer, show an indeterminate `embedded N chunks` state rather than mapping each outer group to 0-100%.

### Suspend causes false stall

Windows power events show sleep/Modern Standby at 19:31:40 and resume at 20:18:54 (~2829 seconds). On resume the guard immediately raised `Indexing made no progress for 2853s`. The executor worker resumed and Ollama continued returning HTTP 200 responses, but the API job had already transitioned to error and could no longer execute the publish step.

Suggested fix: measure no-progress time as accumulated watchdog polling intervals that actually executed, resetting on heartbeat. A long event-loop/suspend gap should not be charged as active no-progress time.

The ownership-token change in this PR correctly suppresses stale progress leakage, but these are separate remaining failure modes. The real-device validation item in the PR description would currently fail on suspend/resume and the progress UI would still report premature 100%.

deeptutor/services/rag/pipelines/llamaindex/pipeline.py
```diff
@@ -79,9 +90,22 @@ async def _run_with_stall_guard(
...
+    # by a stall can no longer write its batch events into the next job
+    # (#1478 — the "infinite Embedding batches / reset to 1/N" leak).
+    owner = f"{time.monotonic_ns()}"
+    global _ACTIVE_PROGRESS_OWNER
+    _ACTIVE_PROGRESS_OWNER = owner
+
     last_progress = {"at": time.monotonic()}
 
     def _heartbeat(*args: Any, **kwargs: Any) -> None:
+        if _ACTIVE_PROGRESS_OWNER != owner:
```
AmirF194, line 105: Ran this against `e0dc389a` (base `897fce52`) in a clean container.

`_heartbeat` (pipeline.py:105) checks its owner token against a single global, `_ACTIVE_PROGRESS_OWNER`, overwritten by whichever run enters `_run_with_stall_guard` last. A second concurrent run permanently suppresses the first run's heartbeat, even while the first run keeps genuinely embedding, contradicting the guard's own docstring: "never fails a job that is making progress."

Simulated two concurrent runs: A calls its own progress callback every 30ms (`stall_timeout=0.3`), B starts 100ms later and runs 1s.

```
head e0dc389a: job_a raises IndexingStallError after 3/40 of its own progress calls
base 897fce52: job_a completes normally, all 40 calls honored
```

A set of active owners (add on entry, discard in `finally`), with `_heartbeat` checking membership instead of equality, would keep the #1478 fix without this regression: a finished run's owner leaves the set, a still-running one stays regardless of how many later runs start.

Not tested against a real concurrent-indexing deployment, only the guard function in isolation.

PR #1504, opened 2026-09-17
feat(learning): persist cross-surface Focus-Check in Question Bank

### LINKED ISSUE #1471: [Feature Request]: Adaptive long-term learning engine for Mastery Path — retention-based spaced repetition, unified learning evidence, and study prioritization

KryptonGao:
## Summary
- Give Mastery Path, Book Focus-Check, and Immersive Reading one shared Question Bank write path (`record_assessment`) with explicit assessment type, result, and provenance.
- Grade Immersive Reading quizzes on the server: store answer keys in `reading_quiz_pending`, accept only `question_id` / `selected_index`, and skip persistence when there is no real session.
- Keep Quiz me usable after a slow LLM call, and mark the learner’s pick vs the reference answer on Question Bank cards (letter keys or stored option text).

Closes #1471 (Phase 2: cross-surface review layer). Complements #1475 (adaptive SRS) without bundling it.

## Test plan
- [x] `python -m pytest tests/reading/test_extension_router.py tests/reading/test_quiz.py tests/api/test_reading_quiz_answers.py tests/api/test_notebook_router.py tests/api/test_book_quiz_attempt_notebook.py deeptutor/learning/tests/test_assessment.py deeptutor/learning/tests/test_mastery_tools.py -q`
- [x] Question Bank card matching tests (`web` node tests for `question-bank-answers` / `question-card-assessment`)
- [x] Computer Use: Quiz me returns 3 questions; local Correct/Incorrect before persist; `POST /quiz/answers` has no `correct_choice_index`; retry upsert stays one card
- [x] Computer Use: Immersive Reading + Focus Check badges; Wrong Only / Needs Review hide Correct and ungraded; reading chip is plain text
- [x] Computer Use: incorrect card shows selected option as Your pick and reference option as Correct

### DIFF
--- a/deeptutor/agents/_shared/tool_composition.py
+++ b/deeptutor/agents/_shared/tool_composition.py
@@ -352,6 +352,23 @@ def user_has_question_bank() -> bool:
         return False
 
 
+def partner_can_record_questions() -> bool:
+    """Whether this turn runs inside a partner conversation.
+
+    Partner turns mount ``question_bank`` even when the bank is still empty:
+    recording the first wrong question is exactly the point (#1244) — the
+    tool's ``record`` action files the mistake the learner just owned up to
+    into the bank their family reviews. The partner context is read lazily
+    (import-time this would be a cycle through the partners package).
+    """
+    try:
+        from deeptutor.services.partners.interaction import get_partner_turn_context
+
+        return get_partner_turn_context() is not None
+    except Exception:
+        return False
+
+
 __all__ = [
     "AUTO_MOUNTED_TOOLS",
     "ToolMountFlags",
@@ -359,6 +376,7 @@ __all__ = [
     "admin_enabled_optional_tools",
     "compose_enabled_tools",
     "default_optional_tools",
+    "partner_can_record_questions",
     "user_has_mastery_topics",
     "user_has_memory",
     "user_has_notebooks",
--- a/deeptutor/agents/loop/pipeline.py
+++ b/deeptutor/agents/loop/pipeline.py
@@ -31,6 +31,7 @@ from deeptutor.agents._shared.tool_composition import (
     ToolMountFlags,
     compose_enabled_tools,
     default_optional_tools,
+    partner_can_record_questions,
     user_has_mastery_topics,
     user_has_memory,
     user_has_notebooks,
@@ -698,7 +699,11 @@ class AgenticLoopPipeline:
                 has_sources=False,
                 has_memory=user_has_memory(),
                 has_notebooks=user_has_notebooks(),
-                has_question_bank=user_has_question_bank(),
+                # A partner turn mounts the bank even when it is still empty:
+                # the tool's record action must be available to file the FIRST
+                # wrong question the learner owns up to (#1244). Product chat
+                # keeps the entries-exist gate.
+                has_question_bank=user_has_question_bank() or partner_can_record_questions(),
                 has_skills=bool(context.skills_manifest),
                 has_deferred_tools=getattr(self, "_deferred_loader", None) is not None,
                 has_exec=getattr(self, "_exec_enabled", False),
--- a/deeptutor/api/routers/question_notebook.py
+++ b/deeptutor/api/routers/question_notebook.py
@@ -18,7 +18,9 @@ from deeptutor.services.storage import get_attachment_store
 logger = logging.getLogger(__name__)
 
 router = APIRouter()
-AssessmentSource = Literal["deep_question", "mastery_path", "immersive_reading", "book"]
+AssessmentSource = Literal[
+    "deep_question", "mastery_path", "immersive_reading", "book", "partner_chat"
+]
 ScoreTrend = Literal["new", "improved", "declined", "unchanged"]
 
 
--- a/deeptutor/services/partners/interaction.py
+++ b/deeptutor/services/partners/interaction.py
@@ -115,6 +115,9 @@ class PartnerTurnContext:
     store: PartnerSessionStore
     own_memory: PathService
     shared_memory: PathService
+    #: Display name of the partner, for surfaces that label content with it
+    #: (e.g. the question-bank session a partner records mistakes into).
+    partner_name: str = ""
 
     @property
     def actor_id(self) -> str | None:
@@ -132,6 +135,7 @@ def build_partner_turn_context(
     store: PartnerSessionStore,
     *,
     legacy_own_memory: PathService,
+    partner_name: str = "",
 ) -> PartnerTurnContext:
     actor_id = personal_actor_id(actor)
     if actor_id is None:
@@ -141,6 +145,7 @@ def build_partner_turn_context(
             store=store,
             own_memory=legacy_own_memory,
             shared_memory=get_admin_path_service(),
+            partner_name=partner_name,
         )
 
     assert actor is not None
@@ -154,6 +159,7 @@ def build_partner_turn_context(
         store=store,
         own_memory=PathService(workspace_root=private_workspace),
         shared_memory=get_path_service_for_scope(actor.scope),
+        partner_name=partner_name,
     )
 
 
--- a/deeptutor/services/partners/runtime.py
+++ b/deeptutor/services/partners/runtime.py
@@ -457,6 +457,7 @@ class PartnerRunner:
                     msg.actor,
                     store,
                     legacy_own_memory=get_current_path_service(),
+                    partner_name=self.config.name,
                 )
                 with partner_turn_context(turn_context):
                     event_stream = get_turn_engine().execute(context)
--- a/deeptutor/services/session/__init__.py
+++ b/deeptutor/services/session/__init__.py
@@ -4,6 +4,7 @@ from .protocol import SessionStoreProtocol
 from .sqlite_store import (
     SQLiteSessionStore,
     get_sqlite_session_store,
+    get_sqlite_session_store_for,
     make_imported_session_id,
 )
 from .turn_runtime import TurnRuntimeManager, get_turn_runtime_manager
@@ -43,6 +44,7 @@ __all__ = [
     "TurnRuntimeManager",
     "get_session_store",
     "get_sqlite_session_store",
+    "get_sqlite_session_store_for",
     "get_turn_runtime_manager",
     "make_imported_session_id",
 ]
--- a/deeptutor/services/session/sqlite_store.py
+++ b/deeptutor/services/session/sqlite_store.py
@@ -96,7 +96,9 @@ def _json_loads(value: str | None, default: Any) -> Any:
 # this id prefix as their discriminator (see ``SQLiteSessionStore._WHERE_*``).
 _IMPORTED_ID_PREFIX = "imported_"
 _ID_SAFE = re.compile(r"[^A-Za-z0-9_-]")
-ASSESSMENT_SOURCES = frozenset({"deep_question", "mastery_path", "immersive_reading", "book"})
+ASSESSMENT_SOURCES = frozenset(
+    {"deep_question", "mastery_path", "immersive_reading", "book", "partner_chat"}
+)
 SCORE_TRENDS = frozenset({"new", "improved", "declined", "unchanged"})
 ACTIVE_TURN_STATUSES = frozenset({"queued", "running", "waiting_input"})
 TERMINAL_TURN_STATUSES = frozenset({"completed", "failed", "cancelled"})
@@ -736,6 +738,33 @@ class SQLiteSessionStore:
         session_id: str | None = None,
     ) -> dict[str, Any]:
         return await self._run(self._create_session_sync, title, session_id)
+
+    async def ensure_notebook_session(self, session_id: str, title: str) -> bool:
+        """Insert a placeholder session row so notebook-only sources can file
+        entries against it (partner chat has no rows in this store otherwise).
+
+        Returns ``True`` when a row was created, ``False`` when it existed.
+        """
+        return await self._run(self._ensure_notebook_session_sync, session_id, title)
+
+    def _ensure_notebook_session_sync(self, session_id: str, title: str) -> bool:
+        resolved_id = (session_id or "").strip()
+        if not resolved_id:
+            raise ValueError("ensure_notebook_session requires a session_id")
+        now = time.time()
+        with self._connect() as conn:
+            exists = conn.execute("SELECT id FROM sessions WHERE id = ?", (resolved_id,)).fetchone()
+            if exists is not None:
+                return False
+            conn.execute(
+                """
+                INSERT INTO sessions (id, title, created_at, updated_at)
+                VALUES (?, ?, ?, ?)
+                """,
+                (resolved_id, (title or "New conversation").strip()[:100], now, now),
+            )
+            conn.commit()
+        return True
 
     def _get_session_sync(self, session_id: str) -> dict[str, Any] | None:
         with self._connect() as conn:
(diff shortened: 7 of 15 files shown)


yepyhun:
Thanks for this follow-up. Reusing the existing Question Bank, introducing a shared assessment writer, and grading Reading answers from server-stored keys are useful improvements. I reviewed the current code at `04058b90`; the observations below are from static review, not an execution of the test suite.

There are two important distinctions to address before this can satisfy the unified learning-evidence goal in #1471.

### 1. Shared assessment storage is not yet shared learner state

[`record_assessment()`](https://github.com/HKUDS/DeepTutor/blob/04058b903e7c3e454103475d6e59cdcfd6762019/deeptutor/learning/assessment.py#L253-L279) writes Question Bank entries. [`_apply_linkage()`](https://github.com/HKUDS/DeepTutor/blob/04058b903e7c3e454103475d6e59cdcfd6762019/deeptutor/learning/assessment.py#L204-L216) explicitly removes Mastery Path/objective linkage for non-mastery sources, and the Reading answer path does not update the retention scheduler.

So a Reading answer about a concept does not currently change that concept's Mastery Path memory state or next review time. Collecting those answers in one place is a useful first step, but the cross-surface feedback loop is still missing.

Suggested next step: define a narrow, trusted mapping from an assessment to an existing objective, then consume eligible assessment events through the same learning-state update path. Leave unmatched or ambiguous assessments unlinked; please do not guess concept identity from text or accept arbitrary client-supplied linkage. Preserve provenance so the learner can see which assessment caused a scheduling change.

### 2. Preserve attempts separately from the latest Question Bank card

Reading allows another answer to the same question, while the notebook's conflict update [replaces the previous answer/result](https://github.com/HKUDS/DeepTutor/blob/04058b903e7c3e454103475d6e59cdcfd6762019/deeptutor/services/session/sqlite_store.py#L2624-L2647). The Reading writer does not supply an incremented attempt count, so the default remains `1`.

A wrong answer followed by a correct answer therefore does not leave a durable record of both attempts and their timestamps. The transient score trend cannot reconstruct that history. This matters because #1471 asks for evidence that can be audited/replayed, and retries should not count as equivalent to unaided first-attempt success.

Please keep the Question Bank card as a latest-state projection, while preserving distinct assessment attempts in an event history. Prefer integrating with the evidence ownership introduced in #1475 rather than creating a second unrelated learning history. Use an attempt identity so a network retry of the same submission is idempotent, while a genuinely new learner attempt remains a new event. Those are different cases even when the question ID is unchanged.

Useful focused acceptance cases would be: wrong -> correct preserves both events; resending the same submission creates no duplicate event; and an explicitly linked Reading assessment updates the intended objective exactly once.

### Scope and maintainability

Please replace `Closes #1471` with a partial-implementation reference, or explicitly keep the parent issue open. This PR supplies a cross-surface collection/review layer; it does not yet complete the shared retention model requested there. A clearly tracked follow-up for linkage and evidence consumption is fine.

As a smaller cleanup, assessment type/result definitions and normalization are duplicated between `assessment.py` and `sqlite_store.py`. One canonical definition with explicit legacy conversion would reduce drift. The `AssessmentRecord` fields `created_at`, `updated_at`, `bookmarked`, and `resolved` are also accepted but not forwarded by `to_notebook_item()`; either honor them or remove/document those inputs.

Most of the large diff is generated API contracts and tests, so reducing the headline line count is not the priority. The higher-value work is preserving evidence and connecting it safely to the retention engine.

### Practical completion criteria for the connected-evidence layer

To clarify what would make this layer complete in practice:

- Wrong -> correct on the same question preserves two ordered attempt events, while the Question Bank card may show the latest state. Retrying the identical network submission is idempotent and creates no duplicate event.
- A Reading/Book assessment with a trusted objective mapping updates the intended learner state and review schedule exactly once. An unmatched or ambiguous assessment is retained with provenance but does not update an objective.
- The canonical evidence event is durable before scheduling depends on it. If updating the learner state fails, processing can retry safely after restart without losing the event or counting it twice.
- The resulting recommendation exposes enough provenance to explain which assessment changed the review schedule.
- The flow remains correct across a process restart: the attempt history, latest Question Bank projection, linked learner state, and next review remain consistent.

Focused end-to-end cases for these behaviors would be more valuable than a broad test expansion. Together with the completion criteria on #1475, these conditions would complete the bounded adaptive-review + connected-evidence feature. They would not by themselves complete the broader #1471 items such as misconception clustering, prerequisites, or cross-course planning, so the parent issue should remain open or those items should be tracked separately.

Issue #1516, opened 2026-09-19
[Bug]:`BwrapBackend` fails on modern Linux (Ubuntu 24.04+) with `bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`, causing Agent/Partner file generation and IM delivery to fail silently


githubhelin:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

[Bug] BwrapBackend fails on modern Linux (Ubuntu 24.04+) with loopback permission denied, blocking sandbox subprocess fallback and partner file delivery`

---

## 🐛 Problem Description

When running DeepTutor on modern Linux distributions (notably **Ubuntu 24.04 LTS / Noble Numbat** and systems enforcing unprivileged user namespace restrictions via AppArmor):

1. An agent, assistant, or IM partner (e.g. connected via WeChat/Weixin, Telegram, Feishu, Slack) is asked to generate deliverable files (such as Word `.docx` documents, spreadsheets, data visualization plots, or code artifacts) and send them to the user.
2. The agent correctly identifies the intent and attempts to run document generation skills (e.g. `docx` skill) via the `exec` sandbox tool.
3. Every execution silently crashes with:
   ```text
   STDERR:
   bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted

   Exit code: 1
   No new or changed artifacts were produced by this call.
   ```
4. Because the sandbox execution crashes, **no files are generated on disk**. The agent then falls back to searching workspace directories, finds nothing, and is forced to apologize to the user:
   > *"由于系统执行沙箱环境限制（出现权限配置错误），代码生成文件未能成功完成...我无法在此环境下直接生成文件并发送给你。"*
5. External IM channels (such as `WeixinChannel`) cannot dispatch any file cards because the media collection pipeline (`_collect_turn_media_files`) finds zero generated artifacts.
6. Crucially, despite `system.json` having `"sandbox_allow_subprocess": true` configured, DeepTutor's `BwrapBackend.health()` reports `(True, "bwrap functional")` due to a flawed exit-code check, preventing automatic fallback to `RestrictedSubprocessBackend`.

---

## 💻 Environment & Runtime

- **DeepTutor Version**: v1.6.7 / v1.6.8
- **OS**: Ubuntu 24.04 LTS (Noble Numbat) / Debian 12+ (Linux kernel 6.8+)
[...]

PR #1520, opened 2026-09-19
fix(runtime): tolerate bwrap loopback denial on modern Linux (#1516)

Frank-zhu0404:
### Description
Tolerate bwrap loopback denial on modern Linux: treat loopback-probe failure as unhealthy, retry once with `--share-net` (sticky), and fall back to `RestrictedSubprocessBackend` when allowed. Prevents silent/hard sandbox failures when RTM_NEWADDR / loopback bind is denied.

### Related Issues
- Closes #1516

### Module(s) Affected
- [x] `services`
- [x] `tests`

### Checklist
- [x] **This PR targets `dev`** (not `main` — see the base-branch selector above).

### Additional Notes
- Head: `510e6313` on `fix/issue-1516-bwrap-loopback`
- Diff: `backends.py` + `service.py` + `tests/services/sandbox/test_sandbox.py` (+181/−5)
- New coverage: loopback probe → unhealthy; retry + sticky `--share-net`; subprocess fallback when allowed; no fallback when disallowed / from unhealthy runner
- Muse review: `cline -P cline-pass -m cline-free/muse-spark-1.3-contributor --thinking xhigh` → **VERDICT=PASS**
- Local full pytest env not fully resolved on this box (uv conflict on markitdown/youtube-transcript-api); tests shipped in-commit and Muse-reviewed


deeptutor/services/sandbox/backends.py
```diff
@@ -264,15 +285,33 @@ async def exec(self, request: ExecRequest) -> ExecResult:
...
         )
 
+    async def exec(self, request: ExecRequest) -> ExecResult:
+        result = await self._run_bwrap(self._build_argv(request), request)
+        if not self._share_net and _is_bwrap_loopback_denied(result):
+            logger.warning(
+                "bwrap loopback setup denied (%s); retrying with --share-net",
+                (result.stderr or "").strip() or f"exit {result.exit_code}",
+            )
+            self._share_net = True
```
AmirF194, line 307: I ran this against `510e6313` in a clean container, simulating the loopback denial the same way this PR's own test does.

Once `_share_net` flips true, `BwrapBackend.level` stays `IsolationLevel.SYSTEM` (backends.py:160 is a class attribute, never touched by the retry at line 295). `spec.py`'s own docstring says SYSTEM is "OS-enforced" and offered to everyone, APPLICATION is admin-opt-in because it "does not OS-isolate". `pipeline.py:664-666` reads exactly that: `if level is IsolationLevel.SYSTEM: ... SYSTEM isolation serves everyone`, granting shell exec to any non-admin user.

So after the first loopback denial (which your own comment says is common on Ubuntu 24.04+), every later exec on the shared `SandboxService` singleton runs with full host network sharing, for every user, but still reports SYSTEM and still grants exec to non-admins.

```
$ python3 repro.py   # BwrapBackend(), fake _run_bwrap matching test_bwrap_retries_with_shared_net_after_loopback_denial
after first exec: _share_net = True result ok: True
second exec argv includes --share-net: True
reported isolation level after network isolation was dropped: IsolationLevel.SYSTEM
```

Would it make sense for `BwrapBackend.level` to reflect the degraded state (e.g. return APPLICATION once `_share_net` is true), so the exec-policy gate in pipeline.py sees it instead of treating the sandbox as still OS-network-isolated?

Frank-zhu0404: Agreed — good catch. `BwrapBackend.level` is now a property: SYSTEM while network-isolated, APPLICATION after sticky `--share-net` so the pipeline exec-policy gate sees the degrade. Covered in `94b8bb6a` (retry test + dedicated level assertion).

deeptutor/services/sandbox/backends.py
```diff
@@ -264,15 +297,33 @@ async def exec(self, request: ExecRequest) -> ExecResult:
...
```
AmirF194, line 307: Ran it: `BwrapBackend()`, set `_share_net = True`, `level` now returns `APPLICATION`. Confirmed in a clean container at `94b8bb6a`. Thanks for the fix.

Issue #1527, opened 2026-09-19
[Bug][High Severity]: Mastery Path corrupts learning state across outline replacement and invalid assessments


yepyhun:
### Do you need to file an issue?

- [x] I have searched the existing issues and this bug is not already filed.
- [x] I believe this is a legitimate bug, not just a question or feature request.

### Describe the bug

Two Mastery Path state-integrity failures can reinforce each other:

1. **A full outline replacement can attach earned progress to a semantically unrelated objective.**
2. **An invalid or incorrectly keyed assessment can reduce mastery and keep the learner on the same objective, with no supported way to defer that objective and continue the route without falsely marking it mastered.**

The common invariant that is currently violated is:

> Learning evidence must remain attached only to the objective and question for which it was created.

#### 1. Progress leaks into a different objective after outline replacement

`mastery_build(mode=replace)` generates module and knowledge-point IDs from their positions:

```text
<path>_m0_kp0
<path>_m0_kp1
...
```

If the learner completes the original first objective and later asks the tutor to redesign the whole outline, a completely different new first objective receives the same positional ID (`m0_kp0`).

`LearningService.replace_modules()` considers any state with an ID present in the new map to be surviving state. It therefore preserves the old objective's:

- `mastery_levels`
- `qualitative_mastery`
- `quiz_attempts`
- `error_records`
- `repetition_states`
- `review_queue`
- `learner_mastery_overrides`
- Feynman evidence/retry state
- potentially the current cursor and pending/active question

A real trace showed all three conditions simultaneously:

```text
old objective content != new objective content
old objective id == new objective id
new objective status == mastered (mastery_source=system)
```

The replacement tool result consequently reported the brand-new first objective as already mastered, even though it had never been taught or assessed.

This is not a UI cache problem: the incorrect state is present in the backend's persisted map/tool result.

Current root cause in `main`:

- `deeptutor/capabilities/mastery/tools.py::_parse_modules()` creates positional IDs.
- `deeptutor/learning/service.py::replace_modules()` retains every objective-scoped record whose ID occurs in the replacement map.
- `replace_modules_for_path()` also preserves the cursor and pending interaction when the positional ID still exists, without checking whether the objective's content changed.

The documented workflow makes this reachable during normal use: for a full redesign the tutor switches to Outline mode and calls `mastery_build(mode=replace)`. The caller is using the intended operation; the unsafe identity/state migration happens in the service.

[...]

PR #1530, opened 2026-09-20
fix(mastery): keep learning evidence bound to objective identity

KryptonGao:
## Summary
- Bind Mastery Path evidence to objective identity instead of positional IDs: full `replace` keeps an ID only when `(name, type)` is unique, otherwise mints a new ID against reserved live/evidence IDs; targeted revise still uses explicit IDs.
- Add `mastery_repair_question` so an invalid or mis-keyed assessment can be voided or re-keyed, then mastery, errors, spaced-repetition, review, and `learning_evidence` are rebuilt together.
- Add `mastery_defer_objective` so a learner can skip an objective for now without recording fake mastery; `mastery_skip_question` stays question-only and the hard gate remains the default.

Fixes #1527

## Test plan
- [ ] Replace an unrelated first objective: the new waypoint is `new`, not `mastered`, and does not inherit attempts/errors/reviews.
- [ ] Reorder unchanged objectives: evidence follows the objective, not its list position.
- [ ] Delete an objective that has a pending question: the question and active interaction are abandoned.
- [ ] Confirm full replace vs targeted revise keep different identity semantics.
- [ ] Void an invalid question and confirm mastery/review/`learning_evidence` no longer include it.
- [ ] Correct a wrong answer key and confirm the attempt is regraded without voiding.
- [ ] Defer an unmastered objective and confirm the route advances without marking it mastered.
- [ ] Confirm `mastery_skip_question` still stays on the same objective.


yepyhun:
@KryptonGao 
First, thank you for taking this on and for the careful, substantive work here. The PR is a meaningful improvement: it moves the mutations into `LearningService`, uses the existing transaction boundary, differentiates full replacement from targeted revision, and includes focused state-integrity tests. I do not think the line count is a concern; a substantial part of the change is test coverage.

Before merging, though, I think the following issues should be addressed so that #1527 is actually closed rather than only improved.

## 1. Full outline replacement can still transfer progress to an unrelated topic

`assign_objective_identities()` identifies a retained objective from the normalized `(name, type)` pair. It does not include the module/topic or any semantic content. A common generic title is enough to preserve an old ID and therefore its mastery, evidence, queued review, and possibly an active/pending interaction.

For example, a completed `concept` named `Basic concepts` under an “Electricity” module can be replaced by a `concept` with the same name under a “Probability” module. Those are not necessarily the same learning objective, but the new one inherits the old state.

For a full outline replacement, please use fresh IDs by default. If preservation is desired, it needs an explicit, validated mapping from old IDs to new IDs (and should verify the relevant content/context), rather than an implicit title match. Reordering genuinely unchanged objectives can still preserve identity through that mapping. Please add a regression test with the same objective name/type in a different module/topic.

## 2. A voided assessment is not represented consistently in downstream views

`QuizAttempt.voided` is respected by mastery/review recomputation, but `objective_report()` still projects every attempt as a normal correct/incorrect result. The frontend therefore can continue to show an invalidated question as a learner mistake and include it in the displayed totals/gate explanation. Separately, `MasteryRepairQuestionTool` synchronizes a void to the question bank as `is_correct=True`, which replaces one inaccurate state with another.

Please give voided attempts an explicit public representation (for example `voided` plus an optional reason), exclude them from scored totals and mastery/gate explanations, and render them as invalidated rather than correct or incorrect. The question bank should store the same distinct state instead of fabricating a correct answer. Add an end-to-end projection/UI-facing test in addition to the existing service-level recomputation test.

## 3. Repair must be idempotent and must remove the exact evidence for that question

The repair code links a `QuizAttempt` to `LearningEvidence` using only the objective ID, assessment type, and a two-second timestamp window. When no match is found, the fallback removes the newest quiz/review evidence for the objective. This can remove evidence belonging to a different question, particularly if an already-voided question is voided again or several assessments happen close together.

Please persist a stable question/attempt/interaction ID on the evidence record and use it for repair/regrade. A repeated `void` request should either be an explicit no-op or return an “already voided” result; it must never modify a different attempt’s evidence. Add tests for repeated void and for two nearby assessments under the same objective.

## 4. Deferral should work for an active due review too

`next_objective()` can select a due review for an already-mastered objective, while `defer_objective()` rejects mastered objectives. Thus a learner who says “move on for now” during a review can be sent back to the same review with no usable defer route.

Please allow deferring a due review without changing the earned mastery, and define its resume behavior. The existing “defer an unmastered objective” test should be complemented by this case.

## 5. The teach-first / fast-foundation requirement remains unimplemented

The PR retains the unconditional `new -> probe` route and the prompt still instructs the tutor to probe first. Repair and defer are useful recovery mechanisms, but they do not meet the requested behavior where a learner asks for concise teaching/foundation before assessment, or for fewer checks while progressing quickly.

Please add a small, explicit learner-intent/pace policy at the existing mastery-policy layer: teach-first should produce a compact explanation/example before the first probe; the default strict gate can remain unchanged; and an explicit preference should affect probe timing/frequency without marking unlearned material as mastered. This needs a policy-level regression test so it does not remain prompt-only behavior.

With these changes, the PR would cover both core failures from #1527: state must stay bound to the actual learning objective, and learner-control/assessment recovery must be truthful across the whole product surface. Without them, this is strong progress but not a complete resolution of the issue.

PR #1532, opened 2026-09-20
feat(voice): verbalize LaTeX in TTS with a Math speak setting

antmikinka:
## Summary

TTS no longer reads `$` as "dollar". Assistant math is prepared on the server before it reaches the voice model.

- **Always:** unwrap `$…$`, `$$…$$`, `\(...\)`, `\[…\]`, and math environments so delimiters are not spoken.
- **Math speak on (default):** verbalize high-frequency TeX (`\frac` → "over", `^2` → "squared", Greek, sums/integrals, inequalities) inside math islands only. Prose like `file_name` and Windows paths are left alone.
- **Math speak off:** keep inner TeX (`E = mc^2`, `\frac{1}{2}`) after stripping `$`.

The preference lives in Settings → Voice (`ui.voice_math_speak`). Chat does not send a per-request flag; `PUT /api/settings/voice-math-speak` persists it and `synthesize_speech` reads it on every TTS call.

Auto-play remains a separate switch (still default off).

## Why a fork PR

This checkout only has read access to `HKUDS/DeepTutor`, so the branch is on `antmikinka/DeepTutor`. CONTRIBUTING says feature PRs target `dev`.

## Tests

- `tests/services/test_voice.py` — delimiter strip, verbalizer, math-speak off, settings-driven synthesis
- `tests/api/test_settings_router.py` — preference persists without freezing other UI defaults

## Notes

- Dollar stripping is not optional. The toggle only controls verbalization.
- Function words are English (`over`, `squared`). A Chinese lexicon is not in this PR.
- The old `/settings/tts` page does not exist on `dev`; playback toggles are on the Voice / TTS settings sections.


antmikinka:
Follow-up after merge, for review as a draft: https://github.com/HKUDS/DeepTutor/pull/1557

Two gaps found while using this on the v1.5.15 port, and they are still on `dev`:

1. Bold markers can still reach TTS. `_EMPHASIS` requires the inner text to start and end with a non-space, so `**bold**` (and bold wrapped around a math island, which we pad with spaces) is spoken as "asterisk". The follow-up unwraps emphasis after math verbalization and drops leftover `*`.
2. Play aloud still opens "Auto-play replies in this conversation?" on the first click and swallows TTS errors. The speaker should only play that reply. Auto-play stays a Settings preference. Failures toast, and Stop cancels the in-flight request.

The button on `dev` is `web/features/chat/messages/ChatMessageList.tsx`, not the old `ChatMessages.tsx`. #1557 targets `dev`.

antmikinka:
@pancacake wanted to add this. I've been using DeepTutor to teach myself mathematics while in university, and hearing replies aloud helps, but the voice was saying "dollar" and "asterisk" through the formulas and bold text. Play aloud was also asking me to turn on auto-play when I only wanted that one reply.

#1532 got the math-speak toggle in. This follow-up is the rest of what I hit while actually studying with it: https://github.com/HKUDS/DeepTutor/pull/1557

PR #1556, opened 2026-09-22
refactor(auth): converge all role whitelists on a single normalize_role, add first set_role tests

polaris-smart:
## Summary

The `Role` literal is `Literal["admin", "user"]`, and today every place that validates or stores a role re-declares its own hardcoded `{"admin", "user"}` set. This PR converges all of those scattered whitelists onto a single source of truth — `VALID_ROLES` + `normalize_role` in `deeptutor/multi_user/models.py` — and adds the first test coverage for `set_role`.

**This is a defensive hardening refactor, not a behavior change.** The two-value role system is already self-consistent across the API and storage layers; the problem it addresses is that the invariant is enforced by N independent hardcoded copies, so any future drift in one layer silently diverges from the others.

## What changes

New in `deeptutor/multi_user/models.py`:

- `VALID_ROLES: frozenset[str]` — the single source of truth for legal role values
- `normalize_role(value, default="user")` — returns the value when legal, otherwise degrades to the default. Least-privilege by construction: an unknown or corrupted role never keeps or gains privileges.

Every role whitelist now delegates to it (audited by grep over the full repo; this is the complete list):

| Location | Before | After |
|---|---|---|
| `deeptutor/multi_user/identity.py` — `_canonical_record` | inline `{"admin", "user"}` | `normalize_role(..., default_role)` |
| `deeptutor/multi_user/identity.py` — `_migrate_legacy_users` | inline set membership check | `normalize_role(..., role)` |
| `deeptutor/multi_user/identity.py` — `load_users` | inline set membership check | `normalize_role(..., role)` |
| `deeptutor/multi_user/identity.py` — `set_role` | inline `{"admin", "user"}` | `VALID_ROLES` |
| `deeptutor/multi_user/context.py` — `user_from_token_payload` | inline set membership check | `normalize_role(...)` |
| `deeptutor/services/auth.py` — `set_role` | inline `("admin", "user")` | `VALID_ROLES` |
| `deeptutor/api/routers/auth.py` — `SetRoleRequest.role_valid` | inline `("admin", "user")` | `VALID_ROLES` |

The `SetRoleRequest` validator is included deliberately: if the API layer kept a hardcoded subset while the storage layer consulted the shared whitelist, the two layers could drift apart again (a 422 for a value the store itself would accept). Pointing both at `VALID_ROLES` makes that class of drift structurally impossible.

## Why this matters

1. **Corrupted-data defense.** Role values are persisted in user-facing JSON stores. Today a hand-edited or corrupted `"role": "superadmin"` entry is caught only by whichever individual whitelist happens to run; after this PR every entry point degrades it fail-soft to the default (least privilege) instead of trusting it.
2. **The signing path is intentionally unguarded.** `create_token` (`deeptutor/services/auth.py:260`) passes `role` through as a plain parameter and `decode_token` (`deeptutor/services/auth.py:286`) only applies a default (`payload.get("role", "user")`, lines 307/339). There is no whitelist at the token layer — the storage/validation layer *is* the only line of defense, which is exactly why it should have a single, auditable enforcement point.
3. **Groundwork for authorization evolution.** Recent authz work in this repo (#1113, guardian-based delegation) shows the access model growing beyond a single boolean. When role values are eventually extended, doing it via one `VALID_ROLES` set means the change is a one-line diff plus tests instead of a cross-layer hunt.

## Zero behavior change

- Legal roles (`"admin"`, `"user"`) pass through unchanged everywhere.
- Unknown values follow the same degrade-to-`"user"` path they follow today; only the error strings of the two `set_role` guards change (now derived from `sorted(VALID_ROLES)` instead of a hardcoded list).
- No endpoints added or removed, no permission changes, no frontend changes.

## Tests

The repo currently has **zero** tests touching `set_role`. This adds `tests/multi_user/test_role_normalization.py` (16 tests: 11 test functions, one of them @parametrize x6) covering the full chain:

- `normalize_role` / `VALID_ROLES` semantics (pass-through, degrade, custom default)
- storage layer: legal roles survive a `set_role` → `load_users` round-trip; unknown values are rejected; corrupted stored roles degrade on load
- token payload mapping: legal roles preserved, unknown values degrade
- API validator: `SetRoleRequest` accepts the two legal roles, rejects everything else

## References

- #1113 — prior large authz extension accepted into this codebase (guardian delegation)
- Follow-up ground work; no new role values are proposed in this PR.


polaris-smart:
CI note: the `Python Tests` failures are pre-existing on `main` and unrelated to this PR. Run [35735683931](https://github.com/HKUDS/DeepTutor/actions/runs/35735683931) (the base commit this branch is cut from) fails the identical single test on all four Python versions:

```
FAILED tests/core/test_capabilities_runtime.py::test_visualize_capability_reuses_chat_loop_and_preserves_attachments
AssertionError: {'temperature': 0.4, 'max_tokens': 16384, ...} != {'temperature': 0.15, 'max_tokens': 16000, ...}
```

This branch changes nothing in the capabilities/chat-loop domain. All other checks (lint, format, import checks on 3.11–3.14 across Linux/macOS/Windows, web tests) are green, and the 16 new role-normalization tests pass on all four Python versions.

